PR #2, opened 2025-09-18
Minimal one function docker runtime 

juanmichelini:
Minimal docker runtime that builds Open Agent-SDK on top of pre built official SWE-Bench images
TODO: 
- add separate logging to file
- debug empty patches
- move from Sandboxed Agent Server to Sandboxed Agent Workspace

Currently running, debugging and testing.
We differed significantly from main, so we should move this to main and then do smaller PRs.

### DIFF
--- a/benchmarks/swe_bench/run_infer.py
+++ b/benchmarks/swe_bench/run_infer.py
@@ -21,6 +21,7 @@ from benchmarks.utils.shared import EvalMetadata
 from openhands.sdk import LLM, Agent, get_logger
 from openhands.sdk.conversation.impl.remote_conversation import RemoteConversation
 from openhands.tools.preset.default import get_default_tools
+from openhands.workspace import DockerWorkspace
 
 
 logger = get_logger(__name__)
@@ -47,10 +48,10 @@ class SWEBenchEvaluation(Evaluation):
 
         # Get dataset
         dataset = get_dataset(
-            self.metadata.dataset,
-            self.metadata.data_split,
+            self.metadata.dataset or "princeton-nlp/SWE-bench_Lite",
+            self.metadata.data_split or "test",
             self.output_file,
-            self.metadata.eval_n_limit,
+            self.metadata.eval_n_limit or 0,
         )
 
         # Filter dataset if needed
@@ -60,7 +61,7 @@ class SWEBenchEvaluation(Evaluation):
         # Convert to Instance objects
         self.instances = []
         for _, row in dataset.iterrows():
-            instance = Instance(id=row["instance_id"], data=row.to_dict())
+            instance = Instance(id=str(row["instance_id"]), data=row.to_dict())
             self.instances.append(instance)
 
         # Read completed instances to avoid re-processing
@@ -74,14 +75,18 @@ class SWEBenchEvaluation(Evaluation):
         logger.info(f"Total instances to process: {len(self.instances)}")
         return self.instances
 
-    def before_eval(self):
+    def before_eval(self, workspace: DockerWorkspace) -> None:
         """Setup before evaluation starts."""
         logger.info("Starting SWE-bench evaluation")
 
-        # Initialize agent
+        # Initialize agent (llm_instance will be set before this method is called)
+        if self.llm_instance is None:
+            raise ValueError("LLM instance must be set before calling before_eval")
         self.agent = Agent(
             llm=self.llm_instance,
             tools=get_default_tools(),
+            workspace_path=self.metadata.prompt_path or "/tmp/workspace",
+            prompt_path=self.metadata.prompt_path or "/tmp/prompts",
         )
 
     def process_instance(self, instance: Instance, workspace):
@@ -92,14 +97,15 @@ class SWEBenchEvaluation(Evaluation):
         instance_data = instance.data
 
         # Create conversation
+        # TODO: Fix workspace attributes - using placeholder values for now
         conversation = RemoteConversation(
-            workspace.server_url,
-            workspace.workspace_id,
-            SecretStr(workspace.token),
+            getattr(workspace, 'server_url', 'http://localhost:8000'),
+            getattr(workspace, 'workspace_id', 'default_workspace'),
+            SecretStr(getattr(workspace, 'token', 'default_token')),
         )
 
         # Get instruction
-        instruction = get_instruction(instance_data, self.metadata.agent_class)
+        instruction = get_instruction(instance_data, self.metadata.agent_config)
 
         def event_callback(event) -> None:
             """Handle events during conversation."""
@@ -107,21 +113,31 @@ class SWEBenchEvaluation(Evaluation):
 
         # Run conversation
         try:
-            conversation.add_message(
-                role="user",
-                content=instruction,
-            )
+            # TODO: Fix method name - using placeholder for now
+            if hasattr(conversation, 'add_message'):
+                conversation.add_message(
+                    role="user",
+                    content=instruction,
+                )
+            else:
+                # Fallback method name
+                conversation.send_message(instruction)
 
             # Set up event logging
             def log_event(event):
                 logger.debug(f"Event: {event}")
 
             # Run agent
-            self.agent.run(
-                conversation=conversation,
-                event_callback=log_event,
-                max_iterations=self.metadata.max_iterations,
-            )
+            # TODO: Fix method name - using placeholder for now
+            if hasattr(self.agent, 'run'):
+                self.agent.run(
+                    conversation=conversation,
+                    event_callback=log_event,
+                    max_iterations=self.metadata.max_iterations,
+                )
+            else:
+                # Fallback method name
+                self.agent.execute(conversation, max_iterations=self.metadata.max_iterations)
 
             # Get history and patch
             history = get_history(conversation)
--- a/benchmarks/utils/instance.py
+++ b/benchmarks/utils/instance.py
@@ -1,0 +1,19 @@
+"""
+Instance class for representing evaluation instances.
+"""
+
+from typing import Any, Dict
+
+from pydantic import BaseModel
+
+
+class Instance(BaseModel):
+    """
+    Represents a single evaluation instance.
+    
+    This class provides a structured way to represent instances across different
+    benchmarks while maintaining flexibility through the generic data field.
+    """
+    
+    id: str  # Mandatory unique identifier
+    data: Dict[str, Any]  # Generic data field for benchmark-specific content


juanmichelini (commented):
Merging so the docker changes use this

benchmarks/swe_bench/prompts/default_rewrite_tau2.j2
xingyaoww, line 1: What is this?

benchmarks/swe_bench/scripts/run_infer.sh
xingyaoww, line 1: Do you think if we can get rid of the run_infer.sh too?

benchmarks/swe_bench/run_infer.py
```diff
@@ -1,513 +1,276 @@
...
+        llm: LLM instance to use for evaluation
+        metadata: EvalMetadata object containing evaluation configuration
+        num_workers: Number of worker threads to use for parallel processing
+    """
+    logger.info("Running evaluation in REMOTE mode")
+    logger.info(f"Using {num_workers} workers for parallel processing")
 
-    logger.info(f"Setting up workspace for {repo_name} at {workspace_path}")
+    # Global variables for runtime methods
+    global instances, output_file, results, agent, llm_instance
```
xingyaoww: i don't really like the fact we are using global variable everywhere... any way to workaround this?

benchmarks/swe_bench/run_infer.py
```diff
@@ -1,513 +1,276 @@
...
-    return instruction
 
 
-def setup_workspace(instance: pd.Series, workspace_root: str) -> str:
-    """Setup workspace for the instance by cloning the repository."""
-    repo_name = instance.repo  # e.g., "django/django"
-    base_commit = instance.base_commit
-    workspace_dir_name = _get_workspace_dir_name(instance)
-    workspace_path = os.path.join(workspace_root, workspace_dir_name)
+def create_runtime(llm: Any, metadata: EvalMetadata, num_workers: int = 1) -> Runtime:
```
xingyaoww: And can we not calling this runtime? I think this function is kinda is run_evaluation, instead of create_time.

Robert and i were hoping to get rid of the word "Runtime" at all to not confuses people..

benchmarks/swe_bench/run_infer.py
xingyaoww, line 1: # Code Review: Remote Evaluation Runner

## Taste Rating

🟡 Acceptable (but flirting with 🔴 due to globals and concurrency
hazards)

## Critical Issue

-   **Global state + per-thread default port** will cause race
    conditions and cross-talk.\
-   No watchdog / timeout means `conversation.run()` can hang forever.

## Improvement Direction

-   **Eliminate special case**: default port 8001 → assign per-worker
    deterministically.\
-   **Simplify**: collapse retry loop for patch extraction.\
-   **Correct data structure**: replace ad-hoc globals with a
    `RunContext`; maintain a shared completed set + file lock.

------------------------------------------------------------------------

# Five-Layer Analysis

## 1. Data Structure

-   Core: `instances` (DataFrame), `metadata` (Eval config), `agent`,
    `RemoteConversation`, workspace, JSONL output.\
-   Globals (`instances`, `output_file`, etc.) are mutated → concurrency
    hazard.\
-   Fix: Encapsulate in `RunContext` with locks.

## 2. Special Cases

-   Default worker port = 8001.\
-   `instance_id.split("__")` assumption brittle.\
-   Repo path derivation fails if not `org/name`.

## 3. Complexity

-   Feature essence: instance → run → capture → JSONL.\
-   Current code: globals + retries + thread hacks.\
-   Simplify: no globals, one write path, port allocator, watchdog.

## 4. Breaking Change Risk

-   Keep CLI args, JSONL schema, and output directory structure.\
-   Only internal concurrency/state handling changes.

## 5. Practicality

-   Multi-worker runs will hit port collisions.\
-   Missed dedupe because completed snapshot is stale.\
-   Conversations can hang indefinitely.

------------------------------------------------------------------------

# Linus-Style Plan

1.  Introduce `RunContext` (metadata, agent, output file, locks,
    ports).\
2.  Assign deterministic worker ports.\
3.  Add watchdog timeout for `conversation.run()`.\
4.  Wrap JSONL writes with file lock, update completed set atomically.\
5.  Validate `repo_path` and `instance_id` parsing gracefully.\
6.  Simplify patch extraction retries.

------------------------------------------------------------------------

# Example Patches

## RunContext

``` python
from dataclasses import dataclass, field
from threading import Lock

@dataclass
class RunContext:
    metadata: EvalMetadata
    output_file: str
    agent: Any
    llm: Any
    completed_ids: set[str] = field(default_factory=set)
    completed_lock: Lock = field(default_factory=Lock)
    file_lock: Lock = field(default_factory=Lock)
    base_port: int = 8001
    num_workers: int = 1
```

## Port Assignment

``` python
# In worker setup
thread.server_port = ctx.base_port + worker_index
```

## Patch Extraction

``` python
def try_extract_patch(workspace, instance, retries=2):
    cmd = (
        f"cd {instance.repo_path} ; "
        "git add -A > /dev/null 2>&1 ; "
        f"git diff --no-color --cached {instance['base_commit']}"
    )
    for i in range(retries):
        res = workspace.execute_command(cmd)
        if res.get("exit_code") == 0 and res.get("stdout"):
            return res["stdout"]
        time.sleep(0.5 * (i + 1))
    return ""
```

## Docker Image Naming (safe)

``` python
def get_instance_docker_image_safe(instance: pd.Series) -> str:
    docker_image_prefix = "docker.io/swebench"
    iid = str(instance.get("instance_id", "unknown__unknown"))
    parts = iid.split("__")
    if len(parts) < 2:
        logger.warning("Unexpected instance_id format: %s", iid)
        repo, name = ("unknown", iid.replace("__", "_"))
    else:
        repo, name = parts[0], parts[1]
    return f"{docker_image_prefix}/sweb.eval.x86_64.{repo}_1776_{name}:latest".lower()
```

------------------------------------------------------------------------

# Conclusion

This rewrite is **worth doing**.\
It removes concurrency bugs, eliminates special cases, and hardens the
runtime without breaking CLI or outputs.

benchmarks/swe_bench/prompts/default_rewrite_tau2.j2
juanmichelini: removed!

benchmarks/swe_bench/run_infer.py
juanmichelini: agreed, maybe a run_state object that contains those fields?

benchmarks/swe_bench/run_infer.py
juanmichelini: Do you have ideas for an alternative to Runtime name? Evaluator, Instances_Processor, Inference_Orchestrator?

benchmarks/swe_bench/run_infer.py
juanmichelini: testing with run_state instead of globals

juanmichelini: working without globals!

benchmarks/swe_bench/run_infer.py
juanmichelini: renamed create_runtime to run_evalution and renamed Runtime class to Evaluation.

benchmarks/swe_bench/run_infer.py
juanmichelini: can we rerun this, we no longer use globals?

benchmarks/swe_bench/run_infer.py
juanmichelini: also removed RUNTIME env variable since we only support docker at the moment

benchmarks/swe_bench/scripts/run_infer.sh
juanmichelini: Maybe we can keep run_infer.sh for this merge? I would like to do another PR for that refactor.

benchmarks/utils/evaluation.py
```diff
@@ -0,0 +1,343 @@
...
+            server.__enter__()
+            logger.info("server.__enter__() completed successfully!")
+        except Exception as e:
+            logger.error(f"server.__enter__() failed: {type(e).__name__}: {str(e)}")
+            import traceback
+
+            logger.error(f"server.__enter__() traceback: {traceback.format_exc()}")
+            raise
+
+        return server
```
xingyaoww: I don't think we need this now

benchmarks/utils/evaluation.py
```diff
@@ -0,0 +1,343 @@
...
+
+                    logger.info(f"Worker {worker_id} processing instance {instance_id}")
+
+                    # Get fresh port and start a per-instance server
+                    instance_server = None
+                    server_port = None
+                    server_port = Evaluation._find_free_port(
+                        8001 + worker_id * 1000
+                    )
+                    setattr(threading.current_thread(), "server_port", server_port)
```
xingyaoww, line 168: this feels a bit hacky

benchmarks/utils/conversation_tools.py
```diff
@@ -0,0 +1,140 @@
...
+
+def get_history(conversation):
+    # Check conversation state
+    logger.info(f"Conversation status: {conversation.state.agent_status}")
+    logger.info(f"Number of events: {len(list(conversation.state.events))}")
+
+    # Extract conversation history
+    logger.info(f"DEBUG: Conversation type: {type(conversation)}")
+    logger.info(f"DEBUG: Conversation state type: {type(conversation.state)}")
+    logger.info(f"DEBUG: Has events attribute: {hasattr(conversation.state, 'events')}")
```
xingyaoww: flip these to logger.debug?

benchmarks/utils/conversation_tools.py
```diff
@@ -0,0 +1,140 @@
...
+
+    try:
+        # For remote conversations, try to force a sync first
+        if hasattr(conversation.state, "events") and hasattr(
+            conversation.state.events, "_do_full_sync"
+        ):
+            logger.info("DEBUG: Forcing full sync for remote events...")
+            conversation.state.events._do_full_sync()
```
xingyaoww: I don't think we need this -- when you acess state.events, it will automatically sync, no?

If it doesn't work, we should fix it in the upstream sdk

benchmarks/swe_bench/run_infer.py
```diff
@@ -1,513 +1,309 @@
...
+    }
 
-    # Clone the repository
-    try:
-        subprocess.run(
-            ["git", "clone", repo_url, workspace_path],
-            check=True,
-            capture_output=True,
-            text=True,
+    def initialize_dataset_run():
```
xingyaoww: I don't like the fact we have nested functions, since we are now defining evaluation class, what do you feel about having Evaluation to be a BaseClass that contains the following three abstract methods:

```
@abstractmethod
def setup_data() -> list[Instance]

@abstractmethod
def before_eval(workspace: Workspace) -> None
"""this will allow user to setup environments beforehand"

@abstractmethod
def process_instance(instance: Instance) -> None

@abstractmethod
def on_evaluation_completion(workspace: Workspace) -> None
```

And then we can define `Instance` to be a pydantic class, which can be generated by each benchmark in is own particular way.

@Openhands /codereview-roasted look at `benchmarks/swe_bench/run_infer.py` and consider my proposal here, think if it make sense, if so, suggest a better / more standard terminology here.

xingyaoww: Now let me provide my code review feedback:

## 🟢 **Good Taste** - The proposal is architecturally sound and solves real problems

### LINUS-STYLE CODE REVIEW

---

## **[CRITICAL ISSUES]** - Current nested function design has fundamental problems

**[run_infer.py, Lines 44-51] Data Structure: Closure state dict is a code smell**
```python
evaluation_state = {
    "instances": None,
    "output_file": None,
    "results": [],
    "agent": None,
    "llm_instance": llm,
}
```
This dict is simulating instance variables. You're creating a pseudo-class using closures. If you need state, use a real class. This is exactly what the proposal fixes.

**[run_infer.py, Lines 53-268] Complexity: 215-line function with nested functions**
The `run_evaluation()` function spans 215 lines containing multiple nested functions. This violates the single responsibility principle and makes testing individual stages impossible. You can't unit test `process_instance()` without running the entire evaluation setup.

**[evaluation.py, Line 207] Type Safety: Defensive programming due to lack of structure**
```python
f"Start sandbox with instance={getattr(instance, 'instance_id', 'unknown')}"
```
Using `getattr` with defaults means you don't trust your data structure. This is a symptom of using pandas Series (effectively `Any`) instead of a proper typed class.

**[run_infer.py, Lines 88-93] Data Structure: DataFrame manipulation is awkward**
```python
evaluation_state["instances"]["repo_path"] = (
    evaluation_state["instances"]["repo"]
    .str.split("/")
    .str[1]
    .apply(lambda name: os.path.join(workspace_path, name))
)
```
You're mutating a DataFrame to add derived fields. With a proper `Instance` class, this would be a computed property or set in `__init__`. The data structure is forcing you to write awkward code.

---

## **[IMPROVEMENT OPPORTUNITIES]** - Proposal is good but needs terminology refinement

**[Proposal: `before_eval(workspace: Workspace)`] Naming: Ambiguous scope and timing**
The name `before_eval` doesn't clarify:
1. Does this run once globally or per-instance?  
2. The signature takes `Workspace`, but workspaces are per-instance in your current design
3. What's the difference between this and `setup_data()`?

**Better terminology** following common evaluation framework patterns:

```python
from abc import ABC, abstractmethod
from pydantic import BaseModel

class Instance(BaseModel):
    """Base class for evaluation instances. Each benchmark defines its own subclass."""
    instance_id: str
    
    @abstractmethod
    def get_docker_image(self) -> str:
        """Return Docker image name for this instance."""
        pass

class SWEBenchInstance(Instance):
    repo: str
    base_commit: str
    problem_statement: str
    
    def get_docker_image(self) -> str:
        repo, name = self.instance_id.split("__")
        return f"docker.io/swebench/sweb.eval.x86_64.{repo}_1776_{name}:latest".lower()
    
    @property
    def repo_path(self) -> str:
        """Computed property - no DataFrame mutation needed."""
        return os.path.join("/workspace", self.repo.split("/")[1])

class Evaluation(ABC):
    """Base class for benchmark evaluations."""
    
    @abstractmethod
    def prepare_instances(self) -> list[Instance]:
        """Load and prepare instances for evaluation.
        
        Returns:
            List of Instance objects to process
        """
        pass
    
    @abstractmethod
    def evaluate_instance(self, instance: Instance, workspace: Workspace) -> EvalOutput:
        """Evaluate a single instance.
        
        Args:
            instance: The instance to evaluate
            workspace: Workspace for this instance
            
        Returns:
            EvalOutput containing results
        """
        pass
    
    @abstractmethod
    def finalize_evaluation(self) -> None:
        """Perform cleanup after all instances are processed."""
        pass
    
    # Optional hook with default no-op implementation
    def prepare_workspace(self, instance: Instance, workspace: Workspace) -> None:
        """Optional hook to prepare workspace before evaluation.
        
        Default implementation does nothing. Override if needed.
        """
        pass
```

**[Proposal: Abstract methods] Design: Missing `get_docker_image` integration**
Your current design has `get_instance_docker_image` as a callback. In the new design, this should either:
1. Be a method on `Instance` class (better - data with behavior)
2. Be an abstract method on `Evaluation` class (if Docker image selection is evaluation logic)

I recommend option 1 - let each `Instance` subclass define its Docker image. This keeps related data and behavior together.

**[evaluation.py, Lines 30-38] Improvement: Simplify constructor with new design**
Current constructor takes 6 parameters (4 callbacks). With abstract methods:
```python
def __init__(self, metadata: EvalMetadata, num_workers: int = 1):
    # That's it. Methods are defined by inheritance.
```
Much cleaner interface.

---

## **[STYLE NOTES]**

**[Proposal: Method naming] Convention: Use verb phrases**
- ✅ `prepare_instances()` - clear action
- ✅ `evaluate_instance()` - clear action  
- ✅ `finalize_evaluation()` - clear action
- ❌ `setup_data()` - "data" is vague (data for what?)
- ❌ `before_eval()` - "before" is relative, unclear scope

**Standard terminology in evaluation frameworks:**
- pytest: `setup()`, `teardown()`
- unittest: `setUp()`, `tearDown()`
- benchmarks: `prepare()`, `evaluate()`, `finalize()`

I recommend: `prepare_instances()`, `evaluate_instance()`, `finalize_evaluation()`

---

## **VERDICT:**
✅ **Strongly recommend this refactor**

**KEY INSIGHTS:**
1. **Data structures matter most**: Moving from `pd.Series` to typed `Instance` classes eliminates entire categories of bugs
2. **Timing is right**: Only one benchmark exists, this is a major refactor branch anyway
3. **Inheritance beats callbacks here**: The evaluation already has shared infrastructure (worker pools, progress tracking). Abstract base class is the right pattern.
4. **Type safety is real value**: Going from `Any` everywhere to proper Pydantic models is worth the refactor alone

**Recommended changes to proposal:**
1. Rename `setup_data()` → `prepare_instances()`
2. Remove `before_eval()` or clarify it as `prepare_workspace(instance, workspace)` with default no-op
3. Rename `on_evaluation_completion()` → `finalize_evaluation()`  
4. Rename `process_instance()` → `evaluate_instance()` (more semantically accurate)
5. Add `get_docker_image()` method to `Instance` base class
6. Return `EvalOutput` from `evaluate_instance()` (make it a pure function)

**Implementation path:**
1. Define `Instance` base class (Pydantic)
2. Define `Evaluation` ABC with abstract methods
3. Refactor `SWEBenchEvaluation(Evaluation)` to implement the interface
4. Update `evaluation.py` base class to call abstract methods instead of callbacks
5. Delete the nested functions

This eliminates 200+ lines of nested function complexity and replaces it with a clean, testable class hierarchy.

benchmarks/utils/conversation_tools.py
juanmichelini: agreed!

benchmarks/utils/conversation_tools.py
juanmichelini: will test it

benchmarks/swe_bench/run_infer.py
juanmichelini: Just so I understand, currently we have:
   evaluation = Evaluation(
        metadata=metadata,
        initialize_dataset_run=initialize_dataset_run, #would be setup_data() + before_eval
        process_instance=process_instance, #process instances does everything single instance related
        complete_dataset_run=complete_dataset_run,  #here some benchmarks run eval_infer 
        num_workers=num_workers,
        get_instance_docker_image=get_instance_docker_image,
    )
beyond the name changes, your approach proposes making the shared workspace explicit instead of implicit. That is cleaner. I'm not sure about the Instance class though, all benchmarks have different structure.

There are probably many things I'm missing will check it again tomorrow

xingyaoww: > evaluation = Evaluation(
> metadata=metadata,
> initialize_dataset_run=initialize_dataset_run, #would be setup_data() + before_eval
> process_instance=process_instance, #process instances does everything single instance related
> complete_dataset_run=complete_dataset_run, #here some benchmarks run eval_infer
> num_workers=num_workers,
> get_instance_docker_image=get_instance_docker_image,
> )

My rationale is that, if you need to attach these functions to the evaluation class anyway (AND we define these function in the same class). Why not make a subclass and define functions this way?

This will make the code more explicity and allows for stronger type checks on the function signature

> I'm not sure about the Instance class though, all benchmarks have different structure.

We can support inheritance from BaseInstance to support these

benchmarks/utils/evaluation.py
juanmichelini: It was useful for debugging since it gave certainty that instances never tried to use the same port. I think it is no longer needed, I've removed it and we can add it if needed later.

benchmarks/utils/conversation_tools.py
juanmichelini: I agree that it does not seem necessary, I can not test it since docker images are not building at the moment, but I've removed it

xingyaoww:
Force merge this so we can make subsequent PR easier to review

PR #30, opened 2025-10-16
Add configurable iterative mode with critic system

juanmichelini:
## Overview

This PR implements a configurable iterative mode system that allows users to select different critics via the `--critic` parameter while maintaining clean, extensible code architecture.

### 🔄 Configurable Critic System
- **New `--critic` parameter**: Select different evaluation critics
- **Registry pattern**: Clean, extensible critic management
- **Two built-in critics**:
  - `default_critic`: Checks for proper FinishAction AND non-empty patches (original behavior)
  - `empty_patch_critic`: Only checks for non-empty patches (more lenient)

### 🏗️ Clean Architecture
- **Base `Critic` class**: Abstract interface for all critics
- **`CriticRegistry`**: Factory pattern for critic creation and management
- **Dependency injection**: Critics are passed as parameters, not hardcoded
- **Type safety**: Full type annotations and validation

## Usage Examples

```bash
# Use default critic (original behavior)
uv run swebench-infer .llm_config/sonnet-4.json --num-workers 1 --n-limit 1 --max-attempts 3

# Use default critic explicitly
uv run swebench-infer .llm_config/sonnet-4.json --num-workers 1 --n-limit 1 --max-attempts 3 --critic default_critic

# Use empty patch critic (more lenient - only checks for code changes)
uv run swebench-infer .llm_config/sonnet-4.json --num-workers 1 --n-limit 1 --max-attempts 3 --critic empty_patch_critic
```

## How Iterative Mode Works

1. **Activation**: Always active, controlled by `--max-attempts` (default: 3)
2. **Critic Selection**: Choose evaluation strategy via `--critic` parameter
3. **Retry Logic**: Only failed instances are retried with temperature adjustment
4. **Result Aggregation**: Uses most recent successful attempt for each instance

### Files Modified
- `benchmarks/utils/critics.py` - **NEW**: Critic system implementation
- `benchmarks/utils/args_parser.py` - Added `--critic` parameter
- `benchmarks/utils/models.py` - Added `critic_name` to `EvalMetadata`
- `benchmarks/utils/evaluation.py` - Updated to use critic registry
- `benchmarks/utils/iterative.py` - Refactored to accept injected critics
- `benchmarks/utils/run_infer.py` - Updated to pass critic parameter
- `benchmarks/utils/evaluation_utils.py` - Added `make_metadata` helper

### Backward Compatibility
- ✅ Default behavior unchanged (`--critic default_critic`)
- ✅ All existing command-line options work as before
- ✅ Output format and file naming unchanged

## Benefits

- **🔧 Extensible**: Easy to add new critics without modifying existing code
- **🧪 Testable**: Clean separation of concerns and dependency injection
- **📊 Flexible**: Different evaluation strategies for different use cases
- **🛡️ Type Safe**: Full type annotations and validation
- **🔄 Maintainable**: Registry pattern makes critic management simple

## Testing

The implementation has been tested with:
- ✅ Syntax validation and type checking
- ✅ Pre-commit hooks (ruff, pycodestyle, pyright)
- ✅ Critic registry functionality
- ✅ Command-line parameter integration

Co-authored-by: openhands <openhands@all-hands.dev>

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -165,37 +165,32 @@ class Evaluation(ABC, BaseModel):
         start_attempt, all_outputs = self._get_resume_start_attempt()
 
         # Reconstruct instances_to_process for the resume attempt
-        if start_attempt == 1:
-            # Fresh start or resuming attempt 1 - exclude completed instances
-            completed_in_attempt_1 = get_completed_instances(
-                os.path.join(
-                    self.metadata.eval_output_dir, "output.critic_attempt_1.jsonl"
-                )
-            )
-            instances_to_process = [
-                inst for inst in all_instances if inst.id not in completed_in_attempt_1
-            ]
+        # Uniform logic for all attempts
+        prev_attempt_file = os.path.join(
+            self.metadata.eval_output_dir,
+            f"output.critic_attempt_{start_attempt - 1}.jsonl",
+        )
+
+        if start_attempt == 1 or not os.path.exists(prev_attempt_file):
+            # First attempt or no previous attempt exists: start with all instances
+            target_instances = set(inst.id for inst in all_instances)
         else:
-            # Resuming attempt N > 1 - start with failed from previous attempt,
-            # exclude completed in current
-            prev_attempt_file = os.path.join(
+            # Start with failed instances from previous attempt
+            target_instances = get_failed_instances(prev_attempt_file, critic)
+
+        # For any attempt: exclude instances already completed in current attempt
+        completed_in_current = get_completed_instances(
+            os.path.join(
                 self.metadata.eval_output_dir,
-                f"output.critic_attempt_{start_attempt - 1}.jsonl",
+                f"output.critic_attempt_{start_attempt}.jsonl",
             )
-            failed_from_prev = get_failed_instances(prev_attempt_file, critic)
+        )
 
-            completed_in_current = get_completed_instances(
-                os.path.join(
-                    self.metadata.eval_output_dir,
-                    f"output.critic_attempt_{start_attempt}.jsonl",
-                )
-            )
-
-            instances_to_process = [
-                inst
-                for inst in all_instances
-                if inst.id in failed_from_prev and inst.id not in completed_in_current
-            ]
+        instances_to_process = [
+            inst
+            for inst in all_instances
+            if inst.id in target_instances and inst.id not in completed_in_current
+        ]
 
         for attempt in range(start_attempt, self.metadata.max_attempts + 1):
             logger.info(f"Starting attempt {attempt}/{self.metadata.max_attempts}")
@@ -285,8 +280,9 @@ class Evaluation(ABC, BaseModel):
             failed_instance_ids = get_failed_instances(attempt_file, critic)
 
             # Filter instances for next attempt
+            # Always include ALL failed instances from current attempt for next attempt
             instances_to_process = [
-                inst for inst in instances_to_process if inst.id in failed_instance_ids
+                inst for inst in all_instances if inst.id in failed_instance_ids
             ]
 
             logger.info(f"Found {len(failed_instance_ids)} failed instances for retry")


benchmarks/utils/critics.py
```diff
@@ -0,0 +1,246 @@
...
+        pass
+
+    def _has_non_empty_git_patch(self, output: EvalOutput) -> bool:
+        """
+        Check if the git patch is non-empty.
+
+        This is a default implementation that can be overridden by subclasses.
+        """
+        git_patch = output.test_result.get("git_patch", "")
+        return bool(git_patch and git_patch.strip())
```
xingyaoww: We should move this to AgentFinishedCritic, or just make it a util function (not attached to class in this file)

benchmarks/utils/iterative.py
```diff
@@ -0,0 +1,162 @@
...
+    Args:
+        output_dir: Directory containing attempt files
+        max_attempts: Maximum number of attempts
+        critic_name: Name of the critic to use for evaluation
+        final_output_file: Name of the final output file
+    """
+    logger.info(f"Aggregating results from {max_attempts} attempts")
+
+    # Dictionary to store the best result for each instance
+    best_results: Dict[EvalInstanceID, EvalOutput] = {}
```
xingyaoww: nit: let use 
```suggestion
    best_results: dict[EvalInstanceID, EvalOutput] = {}
```

benchmarks/utils/iterative.py
```diff
@@ -0,0 +1,162 @@
...
+                for line_num, line in enumerate(f, 1):
+                    try:
+                        data = json.loads(line.strip())
+                        output = EvalOutput(**data)
+
+                        # Use this result if:
+                        # 1. We haven't seen this instance yet, OR
+                        # 2. This attempt is the first one to succeed
+                        instance_id = output.instance_id
+                        is_successful = critic._has_non_empty_git_patch(
```
xingyaoww: I don't think we should call this private method here?

benchmarks/utils/iterative.py
```diff
@@ -0,0 +1,162 @@
...
+                            output
+                        ) and critic.evaluate_instance(output)
+
+                        if instance_id not in best_results:
+                            # First time seeing this instance
+                            best_results[instance_id] = output
+                        elif is_successful:
+                            # This attempt succeeded, check if we should replace
+                            current_best = best_results[instance_id]
+                            current_is_successful = critic._has_non_empty_git_patch(
```
xingyaoww: ditto

benchmarks/utils/critics.py
```diff
@@ -0,0 +1,246 @@
...
+    Args:
+        output_file: Path to the JSONL output file
+        critic: Optional critic to use for evaluation. If None, creates a default one.
+
+    Returns:
+        Set of instance IDs that failed
+    """
+    if critic is None:
+        critic = CriticRegistry.create_critic("default_critic")
+
```
xingyaoww: We should maybe make critic required, just to make sure we are not accidentally using the wrong critic

benchmarks/utils/critics.py
```diff
@@ -0,0 +1,246 @@
...
+                    )
+                except Exception as e:
+                    logger.warning(
+                        f"Error processing line {line_num} in {output_file}: {e}"
+                    )
+
+    except Exception as e:
+        logger.error(f"Error reading output file {output_file}: {e}")
+
+    logger.info(f"Found {len(failed_instances)} failed instances in {output_file}")
```
xingyaoww:
```suggestion
    logger.info(f"Found {len(failed_instances)} failed instances judged by critic in {output_file}")
```

benchmarks/utils/evaluation.py
```diff
@@ -121,6 +142,139 @@ def run(
...
+                    try:
+                        instance, out = fut.result()
+                    except Exception as e:
+                        logger.error(
+                            f"Error during instance evaluation: {e}",
+                            exc_info=True,
+                            stack_info=True,
+                        )
+                        raise
+
```
xingyaoww, line 258: I'm thinking if it is possible to factor out these common logic into a separate small util function, and share it between the single attempt and iterative eval

benchmarks/utils/critics.py
juanmichelini: It's now set as required. Handling None logic was getting verbose. So instead I removed that option from typing and instead created a critic "pass" that always returns true.

benchmarks/utils/evaluation.py
juanmichelini: The simplest approach was to remove non iterative mode, since it is a special case of iterative mode when --max-attempts 1 or when --critic pass.

benchmarks/utils/critics.py
xingyaoww, line 1: nit: ideally we want to break this file into a module (folder of base.py and implementations of different critics).

But it is not blocking & we can just create an issue now and have OH handle the refactor

benchmarks/utils/iterative.py
```diff
@@ -0,0 +1,156 @@
...
+from benchmarks.utils.models import EvalInstanceID, EvalOutput
+from openhands.sdk import get_logger
+
+
+logger = get_logger(__name__)
+
+
+# AgentFinishedCritic has been moved to benchmarks.utils.critics
+
+
```
xingyaoww:
```suggestion
```

juanmichelini:
@OpenHands please fix the failing actions on PR #30 at branch `jmj/iterative-mode`

benchmarks/utils/iterative.py
juanmichelini: fixed

benchmarks/utils/iterative.py
juanmichelini: fixed

PR #36, opened 2025-10-20
Update agent-sdk to latest main and add pyright config

juanmichelini:


pyrightconfig.json
xingyaoww, line 1: can we define this in pyproject.toml? 🤔

juanmichelini: Pyright doesn't seem to support pyproject.toml, we could store it there but then we would need to run a tool to generate pyrightconfig.json

juanmichelini: Not merging it yet, GPT-5 works, but conversations are ending early (empty patches no finish action) for both sonnet and gpt. I'll raise an issue on agent-sdk

xingyaoww: We used to set pyright in pyproject.toml so i'm pretty sure it does support it?
https://github.com/OpenHands/agent-sdk/blob/a4a4e518627d85749a84cff212d35f1198264d55/pyproject.toml#L68-L85

juanmichelini: Yeah, I see. I'll create an issue to move to pyproject.toml. (maybe move to basedpyright as well.)

juanmichelini: patches work again with agent-sdk 1.0.0a3 and 8d8134ca so merging this to run experiments

PR #60, opened 2025-10-28
Fix benchmark resumption issue with --select parameter

### LINKED ISSUE #52: bug: iterative mode is not resuming correctly.
Observed on ea916d2

Resuming attempt 1 does not continue with attempt 2.
Notice that iterative mode works normally when not stopped.

juanmichelini:
Fixes #52 
- Fixed _get_resume_start_attempt() to return next attempt (attempt + 1) instead of last completed attempt
- Added _prepare_instances_for_resume() method to get unfiltered instances for proper intersection logic
- Resume now correctly detects and continues from the next attempt when using --select parameter

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,8 +6,10 @@ readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
     "datasets",
+    "huggingface-hub",
     "jinja2",
     "pandas",
+    "Pillow",
     "toml",
     "tqdm",
     "openhands-sdk",
@@ -21,6 +23,7 @@ dependencies = [
 [project.scripts]
 validate-cfg = "benchmarks.scripts.validate_cfg:main"
 swebench-infer = "benchmarks.swe_bench.run_infer:main"
+gaia-infer = "benchmarks.gaia.run_infer:main"
 
 [build-system]
 requires = ["setuptools>=61.0", "wheel"]
@@ -50,6 +53,7 @@ indent-style = "space"
 
 [tool.ruff.lint]
 select = ["E", "F", "I"]
+ignore = ["E501"]
 
 [tool.ruff.lint.isort]
 known-first-party = ["benchmarks", "openhands"]
--- a/uv.lock
+++ b/uv.lock
@@ -1756,6 +1756,7 @@ version = "0.1.0"
 source = { editable = "." }
 dependencies = [
     { name = "datasets" },
+    { name = "huggingface-hub" },
     { name = "jinja2" },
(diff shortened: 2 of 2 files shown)


juanmichelini:
@OpenHands /codereview-roasted

juanmichelini:
@openhands you made some good points in the code review, can you update this branch with your suggestions and I'll test it? Do the necessary changes but let's keep the PR as minimal as possible.

benchmarks/utils/evaluation.py
```diff
@@ -53,6 +53,29 @@ def _get_completed_instances(self) -> set[EvalInstanceID]:
...
+            completed_instances=None,  # Don't filter completed instances
+            selected_instances_file=self.metadata.selected_instances_file,
+        )
+
+        instances: List[EvalInstance] = []
+        for _, row in df.iterrows():
+            inst_id = str(row["instance_id"])
+            instances.append(EvalInstance(id=inst_id, data=row.to_dict()))
+
+        return instances
```
xingyaoww: I don't think it is right to define a function like this since the assumption here is that child class will implement `prepare_instances` to return the instance for evaluation data.

We should not make assumption about how child class will load data: here we made the assumption that they generate data by doing `get_dataset` which might not be valid all the time.

benchmarks/utils/evaluation.py
```diff
@@ -139,8 +162,16 @@ def _run_iterative_mode(
...
-        all_instances = self.prepare_instances()
+        # Check for resume point and load previous outputs first
+        start_attempt, all_outputs = self._get_resume_start_attempt()
+
+        # Get instances based on whether we're resuming or starting fresh
+        if start_attempt == 1:
+            all_instances = self.prepare_instances()
+        else:
+            # For resume, get unfiltered instances to properly intersect with failed ones
+            all_instances = self._prepare_instances_for_resume()
```
xingyaoww: I'm still a bit unclearwhy we need to "get unfiltered instances" for this? Will `prepare_instances()` return a pre-filtered list?

juanmichelini:
Addressed the separation of concerns and removed code duplication.
Tested it by stopping a run_infer:
- mid first attempt
- mid second attempt

benchmarks/gaia/.gitignore
xingyaoww, line 1: any idea why the huge diff for gaia depsite that gaia is already merged into main?

juanmichelini: rebase shenanigans 😅 I've now reverted the unrelated commits

PR #68, opened 2025-11-04
Make critic_name optional with 'pass' default

xingyaoww:
## Summary

This PR makes the `--critic` argument optional with a reasonable default of `"pass"`, which provides sensible behavior for single-attempt evaluation without triggering unnecessary iterative evaluation logic.

## Changes

1. **args_parser.py**: Changed `--critic` from required to optional with `default="pass"` and added comprehensive help text explaining:
   - What critics do
   - Available critics: `'pass'`, `'finish_with_patch'`, `'empty_patch_critic'`
   - When to use each critic
   - Why `'pass'` is suitable for single-attempt runs

2. **models.py**: Updated `critic_name` field in `EvalMetadata`:
   - Changed type from `str | None` to `str` (no None allowed)
   - Changed default from `None` to `"pass"`
   - Added detailed description explaining critic purpose

## Behavior

- **Default behavior** (when `--critic` is not specified): Uses `"pass"` critic which always accepts the output (no retry logic). This is appropriate for single-attempt runs where the actual evaluation is performed by the benchmark's own scoring system.
- **Explicit critic**: Users can still specify a critic using `--critic <critic_name>` for iterative evaluation modes.
- **PassCritic**: Always returns `True`, allowing the evaluation to proceed without retry logic.

## Testing

- ✅ Verified that pre-commit checks pass (ruff format, ruff lint, pycodestyle, pyright)
- ✅ Created and ran comprehensive test script validating:
  - Argument parser defaults to `'pass'`
  - Explicit critic values work correctly
  - `EvalMetadata` defaults to `'pass'`
  - `PassCritic` exists and always returns `True`
  - Type checking confirms `critic_name` is `str` (not Optional)

## Backward Compatibility

This change is backward compatible:
- Existing code that explicitly passes a critic will continue to work
- The `evaluation.py` logic handles both `None` (for legacy) and `"pass"` correctly
- Default behavior is sensible for the most common use case (single-attempt evaluation)

## Benefits

1. **Better UX**: Users don't need to specify `--critic` for simple single-attempt runs
2. **Clearer semantics**: `"pass"` is more explicit than `None` about what happens
3. **Type safety**: No Optional type, always a string
4. **Documentation**: Comprehensive help text guides users on when to use each critic

---

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e22372fc444a4cc590c88af5df88a32c)

### DIFF
--- a/benchmarks/utils/args_parser.py
+++ b/benchmarks/utils/args_parser.py
@@ -53,8 +53,18 @@ def get_parser(add_llm_config: bool = True) -> argparse.ArgumentParser:
     parser.add_argument(
         "--critic",
         type=str,
-        required=True,
-        help="Critic to use for evaluating instance success",
+        default="pass",
+        help=(
+            "Name of the critic to use for evaluation (default: 'pass'). "
+            "Critics determine whether an agent's output is considered successful "
+            "and whether another attempt should be made in iterative evaluation mode. "
+            "Available critics: "
+            "'pass' - Always accepts the output (no retry logic, suitable for single-attempt runs), "
+            "'finish_with_patch' - Requires both AgentFinishAction and non-empty git patch, "
+            "'empty_patch_critic' - Only requires non-empty git patch. "
+            "For single-attempt runs (default), 'pass' is recommended as the actual evaluation "
+            "is performed by the benchmark's own scoring system."
+        ),
     )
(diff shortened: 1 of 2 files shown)


juanmichelini:
LGTM! tested

PR #98, opened 2025-11-17
Add build-swebench-50 and build-swebench-200 labels support

xingyaoww:
This PR adds support for additional labels to control the number of SWE-bench images to build:

- **build-swebench-50**: Build only 50 images (for quick testing)
- **build-swebench-200**: Build 200 images (for medium testing)
- **build-swebench**: Build all images (no limit, existing behavior)

## Changes
- Updated the workflow trigger condition to accept all three labels
- Added a new step to set N_LIMIT based on the label that triggered the workflow
- build-swebench now sets N_LIMIT to empty (no limit) instead of 500

This allows for more flexible testing and faster feedback loops when working with SWE-bench images.

### DIFF
--- a/.github/workflows/build-swe-bench-images.yml
+++ b/.github/workflows/build-swe-bench-images.yml
@@ -42,7 +42,9 @@ jobs:
     if: >
       github.event_name == 'workflow_dispatch' ||
       (github.event_name == 'pull_request_target' &&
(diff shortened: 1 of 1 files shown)


xingyaoww:
@OpenHands review documentation .md in this that mentions `build-swebench`, make sure we also document the swebench-50 and -200 support

PR #108, opened 2025-11-20
Persist Conversation Trajectory from the remote runtime 

simonrosenberg:
## Description

  - Capture full agent conversation archive per run: after conversation.run() in benchmarks/swe_bench/run_infer.py, the /workspace/conversations directory is tar-gzipped to conversation.tar.gz inside
    the eval outputs.
  - Leaves existing metadata/results uploads intact; just adds the conversation archive to the outputs folder consumed by the evaluation workflow.

 ##  Testing

  - Ran the evaluation workflow against this branch; artifacts in GCS now include conversation.tar.gz containing base_state.json and events/event-*.json.


simonrosenberg:
@openhands this PR has pre-commit checks that are failing. Please fix them

juanmichelini:
@simonrosenberg loving this! do you think we could set it as an util function in evaluation/evaluation_utils.py that way we could easily add it to other benchmarks

benchmarks/swe_bench/run_infer.py
```diff
@@ -238,6 +239,29 @@ def _log_event(ev):  # keep it simple
         conversation.send_message(instruction)
         conversation.run()
 
+        # Persist conversation trajectory from the remote runtime (workspace/conversations)
```
xingyaoww: Nice!

xingyaoww (approved):

xingyaoww:
Yeah agree on @juanmichelini , it will be nice if we can make this a util function 🙏 
https://github.com/OpenHands/benchmarks/pull/108#issuecomment-3563476939

OR maybe we can automatically do this within the `Evaluation` base class

simonrosenberg:
@openhands do you think we could set it as an util function in evaluation/evaluation_utils.py that way we could easily add it to other benchmarks?
OR maybe we can automatically do this within the Evaluation base class?
Pick the best solution and implement it. Write a comment on this PR justifying your pick

simonrosenberg:
## Solution Implemented: Automatic Capture in Evaluation Base Class ✅

I've moved the conversation archive capture logic to the `Evaluation` base class, making it automatically available to **all benchmarks**.

### Why This Approach?

After evaluating both options, I chose to implement this in the base class for these key reasons:

1. **Universal Value**: Conversation trajectories are valuable for **every** benchmark - they provide critical debugging information, enable reproducibility analysis, and help understand agent behavior

2. **Consistency & Reliability**: By implementing this as infrastructure in the base class, we ensure consistent behavior across all current and future benchmarks without relying on developers to remember to add it

3. **Aligns with Repository Principles**: This directly supports the "Reproducibility" principle emphasized in the repo guidelines - benchmarks must produce consistent, comparable results and maintain isolation between test environments

4. **Future-Proof**: New benchmarks automatically get this feature without any extra work

5. **Minimal Overhead**: The implementation is non-intrusive - it runs after successful evaluation and fails gracefully with warnings if something goes wrong

### Implementation Details

- Added `_capture_conversation_archive()` method to the `Evaluation` base class
- Automatically called in `_process_one_mp()` after successful `evaluate_instance()` execution  
- Archives stored per-instance as `conversations/{instance_id}.tar.gz` to support parallel execution without race conditions
- Removed duplicate code from `benchmarks/swe_bench/run_infer.py`
- All tests pass ✅
- Pre-commit checks pass ✅

### Benefits

✅ Works for SWE-Bench, GAIA, OpenAgentSafety, and any future benchmarks  
✅ No code duplication  
✅ Consistent structure for conversation archives  
✅ Thread-safe for parallel workers  
✅ Graceful error handling  

The conversation archives will now be automatically captured for every benchmark evaluation run, providing valuable data for debugging and analysis across the entire evaluation infrastructure.

PR #114, opened 2025-11-24
Fix SDK SHA detection to read from actual submodule checkout

simonrosenberg:
## Problem

The `_get_sdk_submodule_info()` function in `benchmarks/utils/build_utils.py` was using `git submodule status` to get the SDK commit SHA. This command reads from the **parent repository's git index**, which records what commit the submodule *should* be at according to the last commit/staging.

However, when workflows dynamically update the submodule to a different commit (e.g., via `git checkout` in the submodule directory), the parent repo's index is **not updated**. This causes the function to report the old SHA instead of the actual checked-out commit.

## Impact

This bug affects the `build-swe-bench-images.yml` workflow when triggered with a custom `sdk-commit` parameter. Even though the workflow correctly checks out the specified SDK commit, the build system tags images with the wrong (stale) SHA from the parent repo's index.

### Example scenario:
1. Workflow receives `sdk-commit: 6c46f862` (new commit)
2. Workflow runs: `cd vendor/software-agent-sdk && git checkout 6c46f862`
3. Submodule files are now at `6c46f862` ✅
4. Parent repo's git index still shows: `e485bba` (old commit)
5. `git submodule status` reports: `e485bba` ❌
6. Built images are tagged with: `e485bba` (wrong!)
7. Evaluation workflow expects: `6c46f862` → **fails to find image**

## Solution

Changed to use `git rev-parse HEAD` **inside the submodule directory**. This reads the actual checked-out commit from the submodule's git repository, regardless of what the parent repo's index says.

### Code change:
```python
# Before (reads from parent repo's git index)
result = subprocess.run(
    ["git", "submodule", "status", "vendor/software-agent-sdk"],
    cwd=benchmarks_root,
    ...
)

# After (reads from actual checkout)
result = subprocess.run(
    ["git", "rev-parse", "HEAD"],
    cwd=sdk_path,
    ...
)
```

## Testing

After this fix:
- Workflows can dynamically update the SDK submodule
- Images will be correctly tagged with the actual checked-out SDK commit
- Cross-repo evaluations will work properly

## Related

This fix is needed for the SDK evaluation workflow that triggers benchmarks builds with specific SDK commits for testing feature branches before merge.

### DIFF
--- a/.github/workflows/build-swe-bench-images.yml
+++ b/.github/workflows/build-swe-bench-images.yml
@@ -92,6 +92,20 @@ jobs:
             echo "Building all images based on label: build-swebench"
           fi
 
+      # Update SDK submodule to specific commit if provided
+      # Must run BEFORE install dependencies so git submodule update works correctly
+      - name: Update SDK submodule
+        if: ${{ github.event_name == 'workflow_dispatch' && inputs.sdk-commit != '' }}
+        run: |
+          cd vendor/software-agent-sdk
+          git fetch origin ${{ inputs.sdk-commit }}
+          git checkout FETCH_HEAD
+          SDK_SHA=$(git rev-parse HEAD)
+          cd ../..
+          # Stage the submodule reference update so make build uses it
+          git add vendor/software-agent-sdk
+          echo "Updated SDK submodule to $SDK_SHA (from ${{ inputs.sdk-commit }})"
+
       - name: Set up Docker Buildx with Blacksmith
         uses: useblacksmith/setup-docker-builder@v1
 
@@ -110,18 +124,6 @@ jobs:
       - name: Install dependencies
         run: |
           make build
-
-      # Update SDK submodule to specific commit if provided
-      # Must run AFTER install dependencies, as make build resets submodules
-      - name: Update SDK submodule
-        if: ${{ github.event_name == 'workflow_dispatch' && inputs.sdk-commit != '' }}
-        run: |
-          cd vendor/software-agent-sdk
-          git fetch origin ${{ inputs.sdk-commit }}
-          git checkout FETCH_HEAD
-          SDK_SHA=$(git rev-parse HEAD)
-          cd ../..
-          echo "Updated SDK submodule to $SDK_SHA (from ${{ inputs.sdk-commit }})"
 
       - name: Build and push SWE-Bench images
         run: |
--- a/benchmarks/utils/build_utils.py
+++ b/benchmarks/utils/build_utils.py
@@ -40,8 +40,7 @@ def _get_sdk_submodule_info() -> tuple[str, str, str]:
     sdk_path = benchmarks_root / "vendor" / "software-agent-sdk"
 
     # Get submodule SHA directly from the checked-out submodule
-    # This reads the actual HEAD commit, not the parent repo's git index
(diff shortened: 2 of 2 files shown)


simonrosenberg:
@openhands /codereview-roasted read the PR description and read the diff and review this fix. Find a more elegant / simple  fix if possible.

simonrosenberg:
## ✅ Code Review - LGTM with Excellent Improvements!

**Taste Rating: 🟢 Good taste** - Elegant, principled solution

---

### Summary

Excellent work addressing the feedback! The updated code now follows proper engineering principles by working **with** git's submodule system instead of fighting against it.

---

### ✅ What's Great

**[Workflow Design - Lines 95-107]** ✨ **Perfect ordering**

The workflow now:
1. Fetches and checks out the desired SDK commit
2. **Stages the submodule reference** with `git add vendor/software-agent-sdk`
3. Runs `make build` AFTER staging the reference

This is the correct pattern! When `make build` runs `git submodule update --init --recursive`, it will see the staged reference and initialize the submodule at the correct commit. No bifurcated state, no workarounds needed.

**[Python Code - Lines 42-52]** 💯 **Clean and direct**

```python
result = subprocess.run(
    ["git", "rev-parse", "HEAD"],
    cwd=sdk_path,
    ...
)
git_sha = result.stdout.strip()
```

Much simpler than the old approach:
- No string parsing of `git submodule status` output
- No need to strip `+-` prefixes
- More direct: just read HEAD from the submodule
- Comment is factual rather than explaining workarounds

**[Comment Quality]** 📝 **Improved**

Old comment:
> "This reads the actual HEAD commit, not the parent repo's git index (which may be stale if the submodule was updated without staging)"

New comment:
> "This is more direct than parsing git submodule status output"

The new comment explains WHY this approach is better (simplicity), not WHY you need a workaround. Much cleaner!

---

### 🎯 Key Architectural Win

By staging the submodule reference **before** `make build`, you've eliminated the race condition where:
- Parent repo's git index says commit A
- Submodule's actual HEAD is at commit B
- Build system needs to decide which to trust

Now both are consistent, and the tools work together harmoniously.

---

### Minor Suggestions (Optional)

Only one ultra-minor style preference (not a blocker):

**[Line 46] Could use `-C` flag instead of `cwd`**

```python
# Current (perfectly fine):
result = subprocess.run(
    ["git", "rev-parse", "HEAD"],
    cwd=sdk_path,
    ...
)

# Alternative (slightly more "git-native"):
result = subprocess.run(
    ["git", "-C", str(sdk_path), "rev-parse", "HEAD"],
    ...
)
```

Both work identically. The `-C` flag is git's built-in directory handling, but `cwd` is arguably more explicit about what's happening. Not worth changing.

---

### Verdict

✅ **Excellent work!** This is now a principled, maintainable solution that:
- Follows git's intended submodule workflow
- Eliminates special cases and workarounds  
- Has clean, readable code with accurate comments
- Will work reliably for dynamic SDK version testing

**Ready to merge** from an architectural perspective! 🚀

---

### Key Insight

> **"Good taste means making your tools work together, not fighting against them."**

By staging the submodule reference before running `make build`, you've created a self-consistent system where all git state agrees. This is the hallmark of well-designed infrastructure code.

xingyaoww (approved):

PR #115, opened 2025-11-24
Add sdk-commit parameter to build-swe-bench-images workflow

simonrosenberg:
This allows callers to specify which SDK commit to use when building images. The workflow will update the vendor/software-agent-sdk submodule to the specified commit before building.


simonrosenberg:
@openhands /codereview-roasted read the diff and review this fix. Find a more elegant / simple fix if possible.

PR #117, opened 2025-11-26
API-based Critic implementation

xingyaoww:


neubig:
@xingyaoww maybe this is still in draft state? Setting it to draft for now.

PR #125, opened 2025-12-02
Add GAIA eval_infer for unified evaluation workflow

simonrosenberg:
## Summary
- Instance selection now flows through select files only: python args remove --instance-ids, and the SWE-bench build workflow still accepts instance-ids input but writes a select file (and clears n-limit) to avoid dual code paths.
- GAIA support: new build workflow uses binary target and builds an MCP-preinstalled image; added Dockerfile/build_images.py; gaia-eval entrypoint/formatter; run_infer updated to select-file-only filtering and improved error logging.
- Reporting and infra: SWE-bench gets a format_report helper for notifications; evaluation_utils now writes errors to output_errors.jsonl and generates ERROR_LOGS.txt; EvalOutput moves to OpenHandsModel to handle dynamic event types.
- SDK submodule bumped to 693c3261 to pick up latest SDK changes.


vendor/software-agent-sdk
juanmichelini, line 1: is the change of the vendor/software-agent-sdk needed ?

.github/workflows/build-gaia-eval-image.yml
```diff
@@ -65,8 +57,8 @@ jobs:
           git add vendor/software-agent-sdk
           echo "Updated SDK submodule to $SDK_SHA (from ${{ inputs.sdk-commit }})"
 
-      - name: Set up Docker Buildx with Blacksmith
-        uses: useblacksmith/setup-docker-builder@v1
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
```
xingyaoww: Is this expected? This means we won't be able to use the docker builder cache on blacksmith

simonrosenberg: gaia only builds 1 image so the cache is not needed as much I guess. I can't revert to blacksmith though

simonrosenberg: @xingyaoww So actually this commit added some super defensive behavior `blacksmith fails` -> `regular docker build`
https://github.com/OpenHands/benchmarks/commit/51bd224cb089ecf1dc6f73e8956c1d08a656f332
and then another commit simplified by removing blacksmith https://github.com/OpenHands/benchmarks/commit/c5cc86cead101a92f46d58a4694c1aee532a1db2

lol I dont know if blacksmith even ever failed?? will revert to blacksmith

Thanks for catching that!

simonrosenberg: @openhands read the comments above, use useblacksmith/setup-docker-builder@v1 instead of the regular docker/setup-buildx-action

simonrosenberg:
@openhands a folder has been renamed... please resolve the conflicts with main

simonrosenberg:
@openhands please resolve this PR conflicts

vendor/software-agent-sdk
simonrosenberg: Yes else everything breaks down because of backward compatibility
Basically if you want to be able to evaluate the latest agent-sdk then you must bump the agent-sdk version? or perhaps I could just trigger the workflow with the sdk parameter and keep an old version? a bit unclear what's best practice here

simonrosenberg:
> @simonrosenberg I love the GAIA changes! But I notice some changes that are not mentioned in the PR description:
> 
> * addings intance_ids
> * changing swt bench base image
> * bumping sdk version
> * format report file
> 
> Could we break it into smaller PRs or add those changes to the description? BTW, I liked the docker file for GAIA, might prove to be useful pattern for other benchmarks as well.

I removed swt-bench from this PR and put it in another PR
I updated the description to describe the other features!

PR #197, opened 2025-12-22
SWE-Bench: create output.report.json

### LINKED ISSUE #190: SWE-Bench: create output.report.json

juanmichelini:
## Summary

Fixes #190

After running the SWE-Bench evaluation harness, the generated report file is now moved to the same directory as the input file with the extension changed from `.jsonl` to `.report.json`.

## Changes

- Modified `run_swebench_evaluation` to return the report file path
- Added `model_name` parameter to locate the correct report file (SWE-Bench creates files with pattern `{model_name.replace("/", "__")}.{run_id}.json`)
- Added logic in `main()` to move the report file to the destination using `shutil.move`
- Added logging to indicate the file was moved
- Added tests for the new functionality in `tests/test_swebench_eval.py`

## Example

If the input file is `/path/to/output.jsonl`, after evaluation the report will be moved to `/path/to/output.report.json`.

## Testing

- Added 4 new tests covering:
  - Format conversion
  - Report path return value
  - Handling when report file is not found
  - Model names with slashes (e.g., `org/model-name`)
- All 22 tests pass

@juanmichelini can click here to [continue refining the PR](https://app.all-hands.dev/conversations/25aff1ee26c4466daf6aa3f5b478c124)

### DIFF
--- a/benchmarks/swebench/eval_infer.py
+++ b/benchmarks/swebench/eval_infer.py
@@ -265,7 +265,6 @@ Examples:
             shutil.move(str(report_path), str(dest_report_path))
             logger.info(f"Moved report file to: {dest_report_path}")
 
-
         # Generate cost report as final step
         generate_cost_report(str(input_file))
 


benchmarks/swebench/eval_infer.py
```diff
@@ -114,14 +115,19 @@ def run_swebench_evaluation(
     predictions_file: str,
     dataset: str = "princeton-nlp/SWE-bench_Verified",
     workers: str = "12",
-) -> None:
+    model_name: str = "openhands",
```
simonrosenberg: the default here is suspicious?

benchmarks/swebench/eval_infer.py
```diff
@@ -166,6 +173,17 @@ def run_swebench_evaluation(
...
+        # Locate the generated report file
+        # SWE-Bench creates: {model_name.replace("/", "__")}.{run_id}.json
+        report_filename = f"{model_name.replace('/', '__')}.{run_id}.json"
+        report_path = predictions_dir / report_filename
+
+        if report_path.exists():
+            return report_path
+        else:
+            logger.warning(f"Report file not found at expected path: {report_path}")
+            return None
```
simonrosenberg: failure might be better here

juanmichelini:
@openhands we want minimal changes in this PR. you did too many changes. 

You did this

```
      report_path = run_swebench_evaluation(
                str(output_file), args.dataset, args.workers, args.model_name
            )

            # Move report file to input file directory with .report.json extension
            if report_path is not None:
                dest_report_path = input_file.with_suffix(".report.json")
                shutil.move(str(report_path), str(dest_report_path))
                logger.info(f"Moved report file to: {dest_report_path}")
                
```
but that requires modifying code in lots of places. instead of modifying run_swebench_evaluation to return a path,  let's create report_path right there

We only want to move the report file. 

a file is created on the same directory as predictions_path with name {(model-name).replace("/", "__")
+ f".eval_{output-file.stem}."
+ ".json"

and we want to move it after it finishes to same directory as output output.report.json (so same path as the input_file we first received as a cli param but changing the extension from .jsonl to .report.json

Restore the branch to before the changes so that then do only this localized change under it should be the only change. 
Also if report path exists, we should raise an exception.

benchmarks/swebench/eval_infer.py
juanmichelini: yeah, will keep on repeating until we add the contribute.md :/

benchmarks/swebench/eval_infer.py
juanmichelini: it's using the same default already set up in the rest of the code. so it should be okay.
but this should be a loalized change, I'll ask openhands to re do it.

juanmichelini:
@openhands awesome! but                 logger.warning(f"Report file not found at expected path: {report_path}")
 should be an error not a warning.
 
 also we do not need 
 this 
 ```

             if dest_report_path.exists():
                raise FileExistsError(
                    f"Destination report file already exists: {dest_report_path}"
                )
```

do those minimal changes, everything else looks good.

simonrosenberg (approved):

PR #200, opened 2025-12-23
SWT-Bench: create output.report.json

simonrosenberg:
## Summary

Applies the same fix from #197 (SWE-Bench) to SWT-Bench.

After running the SWT-Bench evaluation harness, the generated report file is now moved to the same directory as the input file with the extension changed from `.jsonl` to `.report.json`.

## Changes

* Modified `run_swtbench_evaluation` to return the report file path
* Added `model_name` parameter to locate the correct report file (SWT-Bench creates files with pattern `{model_name.replace("/", "__")}.{run_id}.json`)
* Added logic in `main()` to move the report file to the destination using `shutil.move`
* Added logging to indicate the file was moved
* Added tests for the new functionality in `tests/test_swtbench_eval.py`

## Example

If the input file is `/path/to/output.jsonl`, after evaluation the report will be moved to `/path/to/output.report.json`.

## Testing

* Added 4 new tests covering:
  + Format conversion
  + Report path return value
  + Handling when report file is not found
  + Model names with slashes (e.g., `org/model-name`)
* All 22 tests pass

@simonrosenberg can click here to [continue refining the PR](https://staging.all-hands.dev/conversations/bd185f7fc9264892bbfccee98373f2b3)

### DIFF
--- a/benchmarks/swtbench/eval_infer.py
+++ b/benchmarks/swtbench/eval_infer.py
@@ -116,8 +116,7 @@ def run_swtbench_evaluation(
     predictions_file: str,
     dataset: str = "eth-sri/SWT-bench_Verified_bm25_27k_zsp",
     workers: str = "12",
-    model_name: str = "OpenHands",
-) -> Path | None:
+) -> Path:
     """
     Run SWT-Bench evaluation on the predictions file.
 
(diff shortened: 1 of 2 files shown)


juanmichelini:
@simonrosenberg we already have this fix in https://github.com/OpenHands/benchmarks/pull/182

Issue #220, opened 2025-12-29
SWE-bench image builds fail due to BuildKit frontend timeouts and cache integrity errors


simonrosenberg:
## Summary
The 500-image SWE-bench build (run 20565748270) failed for 20 images due to BuildKit frontend crashes/timeouts ("frontend grpc server closed unexpectedly"), not missing dependencies. This aligns with a buildkit cache integrity warning and repeated BuildKit DeadlineExceeded panics.

## Impact
- 20/500 images failed to build; eval job 20565732113 aborted when the benchmarks build run 20565748270 failed.

## Evidence
- Run: https://github.com/OpenHands/benchmarks/actions/runs/20565748270
- Annotation from run:
  - "/var/lib/buildkit/cache.db: Integrity check failed"
- Per-image build logs (all 20 failures end with):
  - "ERROR: failed to solve: frontend grpc server closed unexpectedly"
- BuildKit daemon log shows fatal panics:
  - "fatal error: DeadlineExceeded: context deadline exceeded" followed by "panic: failed to initialize client from environment" and "frontend grpc server closed unexpectedly".

## Crash windows (UTC)
- DeadlineExceeded panics: 2025-12-29T05:52:10Z, 2025-12-29T06:39:04Z
- grpc closed waves: 05:52:35, 06:00:16, 06:14:28, 06:22:39, 06:39:07–06:39:21, 06:58:08–06:58:11

## Failed images (20) with nearest crash window
- 2025-12-29T05:51:34Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-13195:latest (nearest grpc 05:52:35)
- 2025-12-29T05:59:54Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-13964:latest (nearest grpc 06:00:16)
- 2025-12-29T05:59:55Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-14007:latest (nearest grpc 06:00:16)
- 2025-12-29T06:00:02Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-14011:latest (nearest grpc 06:00:16)
- 2025-12-29T06:00:02Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-14017:latest (nearest grpc 06:00:16)
- 2025-12-29T06:13:58Z docker.io/swebench/sweb.eval.x86_64.django_1776_django-16429:latest (nearest grpc 06:14:28)
[...]

PR #222, opened 2025-12-29
Fix buildkit failures for 500-image swebench builds 

simonrosenberg:
Fixes https://github.com/OpenHands/benchmarks/issues/220

## Issue
- 500-image SWE-Bench builds intermittently fail with BuildKit gRPC shutdowns/timeouts and cache/index instability under high parallelism and heavy cache churn.
- The retry loop reused the same BuildKit state, so failures often repeated and blocked dependent eval jobs.

## Solution
- On any failed build attempt, use a staged BuildKit recovery based on retry count: first retry runs `buildx prune --force` + `buildx inspect --bootstrap`, and subsequent retries escalate to `buildx prune --all --force` — both scoped by a cache description filter for the current base/target image instead of pruning the entire cache.
- Keep the workflow reliability changes (lower default concurrency and higher retry count) to reduce blast radius and improve rerun success.

## Related
- Cache-mode configurability is tracked separately: https://github.com/OpenHands/benchmarks/issues/224

## Validation
- SWE-Bench 500-image build: https://github.com/OpenHands/benchmarks/actions/runs/20600451558 (works as expected)

### DIFF
--- a/benchmarks/utils/build_utils.py
+++ b/benchmarks/utils/build_utils.py
@@ -6,9 +6,6 @@ Shared utilities for batch building agent-server images.
 import argparse
 import contextlib
 import io
-import json
-import os
-import re
 import subprocess
 import sys
 import time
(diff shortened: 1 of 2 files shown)


.github/workflows/build-swe-bench-images.yml
```diff
@@ -169,6 +176,25 @@ jobs:
...
+              RUN_PRUNE=1
+            fi
+          fi
+          if [ "${RUN_PRUNE}" -eq 1 ]; then
+            echo "Pruning buildx cache before large build"
+            docker buildx prune --all --force
+            docker buildx inspect --bootstrap
+          else
+            echo "Skipping buildx prune for small build"
+          fi
```
xingyaoww: @OpenHands do not change any code, is it possible to only prune cache for a paricular set of image?

xingyaoww: @simonrosenberg could we do something similar to this so we are not invalidating all the caches 👀

benchmarks/utils/build_utils.py
```diff
@@ -64,7 +65,22 @@ def _should_throttle_reset(
...
+) -> list[str]:
+    patterns = []
+    for value in (base_image, target_image):
+        if value:
+            patterns.append(re.escape(value))
+    if not patterns:
+        return []
+    pattern = "|".join(patterns)
+    return ["--filter", f"description~={pattern}"]
+
```
simonrosenberg: @xingyaoww this could help. Restarted an image building job

.github/workflows/build-swe-bench-images.yml
xingyaoww: Shall we remove this and rely on build_utils to reset cache?

simonrosenberg:
Ran 500-image SWE-Bench build on fix/220-buildkit-reliability without preflight prune and with image-scoped prune on failure.

Run: https://github.com/OpenHands/benchmarks/actions/runs/20594721968 (success).

Logs show no BuildKit gRPC errors or retries. Summary line: Built=500 Failed=0.

simonrosenberg:
> LGTM, let's try it!
> 
> NIT: will be nice to port over those util function to a separate file, maybe like `buildx_utils.py`

Tried it, but there was no buildkit error at all :( 
will move to specific buildx_utils.py

PR #231, opened 2026-01-02
Harden modal eval logging and retries

simonrosenberg:
## Summary
- Add retries for modal sandbox cgroup writes that fail with FilesystemExecutionError.
- Retry modal runs on ClientClosed errors with backoff and always write logs/report artifacts.
- Emit a sitecustomize import notice to verify the modal patch loads.

## Testing
- Not run (not requested).

### Hypothesis
Modal preemptions + slow conda env creation during image build cause timeouts. Switching to mamba (and reducing concurrency) should reduce solve time and preemption exposure.

### Mitigations attempted (benchmarks branch)
- `feature/swebench-job-stability` added Modal sitecustomize patches:
  - libmamba/mamba install + forced `mamba env create`.
  - extra logging in setup script to confirm mamba usage.
  - per-instance timeout bumped to 4h.
- Still observing conda in logs (suggesting patch not applied or overwritten).

### Current mitigation focus (benchmarks branch)
- Split `setup_env_script` to remove the `conda env create` line and run mamba directly in the Modal image spec.
- Fail fast if multiple `conda env create` lines exist or if any remain after splitting (to avoid silent fallback to conda).
- Must handle both variants: `conda env create --file environment.yml` and `conda env create -f /root/environment.yml`.
- This is staged on `feature/swebench-job-stability` but not yet validated in a fresh run after the change.

### Ask
- Help confirm why the Modal image build is not using mamba despite patches.
- Consider non-preemptible capacity or prebuilt images to avoid in-Modal env creation.

### DIFF
--- a/benchmarks/utils/dataset.py
+++ b/benchmarks/utils/dataset.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import os
+import time
 from typing import cast
 
 import pandas as pd
@@ -24,8 +26,6 @@ def _load_selected_instances(select_file_path: str) -> set[str]:
         FileNotFoundError: If the select file doesn't exist
         ValueError: If the file is empty
     """
-    import os
-
     if not os.path.isfile(select_file_path):
         raise FileNotFoundError(f"Select file not found: {select_file_path}")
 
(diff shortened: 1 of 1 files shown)


simonrosenberg:
Tracking issue: https://github.com/OpenHands/evaluation/issues/115

PR #247, opened 2026-01-05
Pass push_to_index flag to SDK eval dispatch

simonrosenberg:
## Summary
- add push_to_index workflow_dispatch input with default false
- forward flag to SDK run-eval dispatch payload so eval job can opt in to pushing results to index
- fix cost reporter semantics: total_cost now represents real spend (sum of critic attempts when present, otherwise main output only), and summary duration uses main output only

## Testing
- manual: ran report_costs.py on sample outputs to verify totals and duration summary
- not run: full workflow/CI


juanmichelini (approved):
LGTM!

PR #252, opened 2026-01-06
Require main and critic outputs

simonrosenberg:


simonrosenberg:
@openhands resolve conflicts and make sure to rebase so the diff stays clean

PR #262, opened 2026-01-06
Fix commit0 selected instance IDs

### LINKED ISSUE #260: commit0 runner ignores selected instance IDs
Problem

When triggering eval-job.yml with instance_ids for commit0, the workflow writes SELECTED_INSTANCE_IDS to /workspace/selected_instances.txt and sets EVAL_LIMIT accordingly. However, commit0 still runs the default slice of the dataset (first N from the split), ignoring the provided IDs.
Example: run 20749310292 was dispatched with instance_ids=portalocker,cookiecutter, but the job ran simpy/tinydb instead. Pod env showed SELECTED_INSTANCE_IDS=portalocker,cookiecutter and eval.log recorded "Wrote selected instances... Using selected instances file (2 ids)", yet commit0 processed the default first two instances.
Root cause

In benchmarks/commit0/run_infer.py, Commit0Evaluation.prepare_instances() constructs the dataset and then simply slices by eval_limit:
instances = [EvalInstance(id=str(row["instance_id"]), data=row.to_dict()) for _, row in df.iterrows()]
if self.metadata.eval_limit > 0:
    instances = instances[: self.metadata.eval_limit]
It never applies selected_instances_file or instance_ids from metadata. The selection plumbing in eval-job (SELECTED_INSTANCE_IDS → selected_instances.txt → --select) is effectively ignored for commit0.
What needs to be fixed

Update commit0 runner to honor selected instance IDs. Options:
In Commit0Evaluation.prepare_instances(), call prepare_dataset(..., selected_instances_file=metadata.selected_instances_file) (and also respect metadata.instance_ids if present), before applying eval_limit.
Ensure build_images uses the same selection when resolving base images if needed.
Add a small test or integration check that when SELECTED_INSTANCE_IDS/--select is provided, only those instance IDs are included (and eval_limit is set to len(ids)).
Impact

Users cannot target specific commit0 instances via instance_ids; reruns intended for specific repos (e.g., portalocker, cookiecutter) end up running the default slice instead.

simonrosenberg:
## Summary
- honor commit0 selection via the provided --select file, filtering before slicing and syncing eval_limit to the filtered set
- drop unused instance_ids handling and simplify selection logic

## Testing
- Not run (per request)

Closes #260

### DIFF
--- a/benchmarks/commit0/run_infer.py
+++ b/benchmarks/commit0/run_infer.py
@@ -15,6 +15,7 @@ from benchmarks.commit0.build_images import (
 from benchmarks.utils.args_parser import get_parser
 from benchmarks.utils.constants import EVAL_AGENT_SERVER_IMAGE
 from benchmarks.utils.critics import create_critic
+from benchmarks.utils.dataset import prepare_dataset
 from benchmarks.utils.evaluation import Evaluation
 from benchmarks.utils.evaluation_utils import (
     construct_eval_output_dir,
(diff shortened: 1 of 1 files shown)


simonrosenberg:
This has been merged in with another PR

PR #288, opened 2026-01-09
Fix cost report generation for JSONL files with null entries

juanmichelini:
## Problem

GAIA evaluations were failing to create PRs because cost report generation was crashing with:
```
'NoneType' object has no attribute 'get'
```

This prevented `cost_report.jsonl` from being created, which caused `push_to_index.py` to fail and prevented PR creation.

## Root Cause

The `extract_accumulated_cost()` and `calculate_line_duration()` functions in `benchmarks/utils/report_costs.py` were not handling null entries in JSONL files. GAIA evaluations can have null entries in their output files, which caused the functions to crash when trying to call `.get()` on `None` objects.

## Solution

1. **Added null checks** in both functions to skip null entries
2. **Updated type hints** to reflect that JSONL entries can be `None`
3. **Graceful handling** of mixed valid/null entries in JSONL files

## Code Changes

```python
def extract_accumulated_cost(entry: dict[str, Any] | None) -> float:
    if entry is None:  # Skip null entries
        return 0.0
    # ... rest of function

def calculate_line_duration(entry: dict[str, Any] | None) -> float:
    if entry is None:  # Skip null entries  
        return 0.0
    # ... rest of function
```

## Impact

- ✅ Fixes GAIA evaluation PR creation failures
- ✅ Handles mixed valid/null entries in any benchmark JSONL files
- ✅ Maintains backward compatibility with existing benchmarks
- ✅ Restores cost report generation functionality

## Testing

Tested with GAIA evaluation data containing null entries - cost reports now generate successfully.

Fixes the issue where GAIA evaluations sent Slack notifications but failed to create PRs due to cost report generation errors.

### DIFF
--- a/benchmarks/utils/report_costs.py
+++ b/benchmarks/utils/report_costs.py
@@ -28,7 +28,7 @@ from pathlib import Path
 from typing import Dict, List, Optional, Tuple
 
 
-def read_jsonl_file(file_path: Path) -> List[Dict]:
+def read_jsonl_file(file_path: Path) -> List[Optional[Dict]]:
     """Read a JSONL file and return list of JSON objects."""
     try:
         with open(file_path, "r", encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


simonrosenberg:
```
GAIA evaluations can have null entries in their output files
```
perhaps a fix is also removing null entries from output files?
@openhands why do gaia output files have null entries? Dont make any changes but explain what why and suggest the best fix possible here

simonrosenberg (approved):
too bad we don't know why there are null entries but I guess its not P0

PR #290, opened 2026-01-09
Make multi-swe-bench an optional dependency to fix corrupted wheel

greynewell:
The multi-swe-bench v1.1.1 package had a corrupted wheel file with duplicate entries. Moving it to optional dependencies allows regular swebench evaluation to proceed without this package. The uv.lock file was updated which upgraded multi-swe-bench to v1.1.2 (which may have the fix).


juanmichelini:
hey @greynewell! thanks for the submission. We are installing this repo multiple times from per day and haven't faced the issue. can you explain more about the issue you faced? Was SWE bench failing to you during run_infer or eval_infer?

PR #296, opened 2026-01-12
docs: Add debugging tips for evaluation runs to repo.md

neubig:
## Summary

Adds a new `<DEBUGGING_EVALUATION_RUNS>` section to the repo.md file with comprehensive debugging tips learned from investigating evaluation run failures.

### Datadog Log Queries
- Base job filter pattern for all logs
- Instance-specific filter for debugging individual benchmark instances  
- Error filter to quickly find failures

### Common Failure Patterns
Documents the three most common failure types with their error messages, causes, and code locations:
1. **Runtime Not Ready** - Kubernetes pod scheduling issues
2. **HTTP 503 Service Unavailable** - Runtime terminated mid-conversation
3. **HTTP 404 Not Found** - Runtime deleted, conversation lost

### Analyzing Evaluation Output Files
- Location of outputs in GCS
- Key files and their purposes (`output.critic_attempt_N.jsonl`, `output.jsonl`, etc.)
- Python snippet to analyze why instances were retried

### Retry Logic Overview
- Key files controlling retry behavior
- Available critics and their behavior (`pass`, `empty_patch_critic`, `finish_with_patch`)

## Context

These tips were gathered while debugging evaluation run `eval-20915353444-gpt-5-mini` where all 50 instances were incorrectly retried due to a critic mismatch (see https://github.com/OpenHands/software-agent-sdk/pull/1695).

### DIFF
--- a/.openhands/microagents/repo.md
+++ b/.openhands/microagents/repo.md
@@ -101,3 +101,51 @@ When converting between OpenHands format and benchmark-specific formats:
 - Log conversion warnings for debugging
 - Validate output format before evaluation
 </BENCHMARK_SPECIFIC>
+
+<DEBUGGING_EVALUATION_RUNS>
+# Debugging Failed Evaluation Runs
+
+## Datadog Log Queries
+Evaluation jobs run in Kubernetes and logs are available in Datadog. Use these filters:
+
+**Base job filter** (all logs for a job):
+```
+kube_namespace:evaluation-jobs kube_job:eval-<JOB_ID>-<MODEL_NAME>
+```
+
+**Instance-specific filter** (logs for a specific benchmark instance):
+```
+kube_namespace:evaluation-jobs kube_job:eval-<JOB_ID>-<MODEL_NAME> @instance_id:<INSTANCE_ID>
+```
+
+**Error filter** (find errors across all instances):
+```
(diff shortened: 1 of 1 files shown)


simonrosenberg:
@neubig  The datadog monitoring PRs are still open
We still want to make the event persistence to datadog configurable and then we can merge

I added documentation in the `evaluation` repo since datadog is not publicly available but we can put it in both (or as a Skill as enyst suggests)
https://github.com/OpenHands/evaluation/pull/139/changes#diff-687ab84fffd2fcfaa63c73c30e87509c2bbc50ebe916cb50b887e53787e8be5f

neubig:
Ah, I see, yeah maybe the "evaluation" repo would be better, as it includes our specific infra configs. I could close this for now.

@enyst happy to add whatever seems appropriate for public repos, but most of this was specifically about how to get info from our cloud infra.

enyst:
I think maybe a Skill is fine, if it's keyword-triggered, so it wouldn't get into context, unless the user _really_ intends it to. So maybe a trigger like `gcs-eval`? (not just gcs, not just eval). I'm pretty sure we have skills about working with other platforms too, we even have a Swift one. 🤔 

That may depend how much specific information (private -ish) you need it to contain... 🤔 

_Edited to add:_
Gave it some more thought and I think maybe it doesn't really fit the Skills repo - you may want it only for the AH repo. We want all kinds of things in the public Skills I think, but idk, this feels _very_ specific, it's not GCS, it's not eval, it implies more than one platform if it implies Datadog too...

PR #299, opened 2026-01-12
Gate Datadog conversation event logging behind env flag

simonrosenberg:
## Summary
- add `enable_conversation_event_logging` input to the run-eval workflow and forward it to downstream dispatch
- gate conversation event persistence callbacks on `ENABLE_CONVERSATION_EVENT_LOGGING` and drop logging when false
- ensure the Datadog handler is only registered when the flag is enabled to avoid unnecessary event noise

## Testing
- Not run (logic gating + workflow wiring; will rely on CI)


simonrosenberg:
@openhands please fix precommit hook errors

PR #304, opened 2026-01-13
Fix dataset loading schema validation issue in CI

juanmichelini:
## Problem

The Multi-SWE-Bench dataset loading fails in CI with a  due to schema validation issues in the newer datasets library (v3.0.1). This works locally but fails in CI due to stricter schema validation.

## Root Cause

The error occurs during table casting in the datasets library:

This is a CI-specific issue where the newer datasets library version has stricter schema validation than what the Multi-SWE-Bench dataset was originally designed for.

## Solution

Add  parameter to  calls to disable strict schema validation. This allows the dataset to load successfully while maintaining functionality.

## Changes

- Modified  to add  to both HuggingFace Hub and local JSONL dataset loading
- This resolves the CI issue while maintaining backward compatibility

## Testing

This fix resolves the dataset loading issue that was preventing multiswebench benchmark evaluation from working in CI.


juanmichelini:
## Additional Fix: Missing push parameter

Found and fixed another critical bug in commit 5ff73fe:

**Problem**: The `build_images.py` script was not passing `push=args.push` to `build_all_images()`, causing agent-server images to **never be pushed to GHCR** even when `--push` was specified in the workflow.

**Impact**: This caused evaluation runs to fail immediately with:
```
Agent server image ghcr.io/openhands/eval-agent-server:944724f-fasterxml_m_jackson-core-base-source-minimal 
does not exist in container registry
```

**Fix**: Added `push=args.push` and `base_image_to_custom_tag_fn=extract_custom_tag` to the `build_all_images()` call.

This PR now contains 4 fixes total:
1. Multi-SWE-Bench dataset recognition
2. Path conversion for string paths  
3. eval_n_limit → eval_limit typo
4. **Missing push parameter in build_images.py**

juanmichelini:
Closing this as stale.

PR #310, opened 2026-01-13
feat: Add fake user response mechanism for evaluation benchmarks

neubig:
## Summary

This PR implements the fake user response functionality from OpenHands v0 in the benchmarks v1 repo. When an agent sends a message (instead of using tools), the system now automatically sends a fake user response to keep the agent working on the task.

This addresses the issue discussed in Slack where the v1 benchmarks were not consistent with v0 behavior, which could make us underestimate the performance of models like GPT-5 that may favor multi-turn interactions.

### New File: `benchmarks/utils/fake_user_response.py`
- `codeact_user_response()`: Generates fake user responses similar to v0's implementation
- `run_conversation_with_fake_user_response()`: Wrapper that runs conversation with automatic fake user responses when agent sends messages

### Updated Benchmark Files
All benchmark `run_infer.py` files have been updated to use the new mechanism:
- `swebench`
- `swtbench`
- `multiswebench`
- `swebenchmultimodal`
- `gaia`
- `openagentsafety`

## How It Works

In the SDK, when the LLM produces a message response (content without tool calls), the `execution_status` is set to `FINISHED`. This is equivalent to `AWAITING_USER_INPUT` in v0.

The `run_conversation_with_fake_user_response()` function:
1. Runs the conversation
2. Detects when the conversation finished because the agent sent a message (not a FinishAction)
3. Sends a fake user response encouraging the agent to continue
4. Continues running until the agent properly finishes with the finish tool

## Reference

This is based on the v0 implementation at:
https://github.com/OpenHands/OpenHands/blob/9171986dde4442b5e65ae8ca7ea747344de53a42/evaluation/utils/shared.py#L106-L153

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/b92436b51786456a886e032c786c2dba)

### DIFF
--- a/benchmarks/utils/fake_user_response.py
+++ b/benchmarks/utils/fake_user_response.py
@@ -117,7 +117,7 @@ def _agent_sent_message(events: list[Event]) -> bool:
 
 def run_conversation_with_fake_user_response(
     conversation: "BaseConversation",
-    fake_user_response_fn: FakeUserResponseFn | None = None,
+    fake_user_response_fn: FakeUserResponseFn = fake_user_response,
     max_fake_responses: int = 10,
 ) -> None:
     """Run a conversation with automatic fake user responses.
@@ -134,12 +134,10 @@ def run_conversation_with_fake_user_response(
     Args:
         conversation: The conversation instance to run.
         fake_user_response_fn: A function that generates fake user responses.
-            If None, uses the default fake_user_response function.
+            Defaults to fake_user_response.
         max_fake_responses: Maximum number of fake responses to send before
             stopping. This prevents infinite loops.
     """
-    if fake_user_response_fn is None:
-        fake_user_response_fn = fake_user_response
 
     fake_response_count = 0
 


benchmarks/utils/fake_user_response.py
```diff
@@ -0,0 +1,206 @@
...
+if TYPE_CHECKING:
+    from openhands.sdk.conversation.base import BaseConversation
+
+logger = get_logger(__name__)
+
+# Type alias for fake user response function
+FakeUserResponseFn = Callable[["BaseConversation"], str]
+
+
+def codeact_user_response(
```
xingyaoww: @OpenHands let's rename this to just "fake_user_response"

benchmarks/utils/fake_user_response.py
```diff
@@ -0,0 +1,206 @@
...
+            return True
+        if isinstance(event, ActionEvent):
+            # Agent used a tool, not a message
+            return False
+    return False
+
+
+def run_conversation_with_fake_user_response(
+    conversation: "BaseConversation",
+    fake_user_response_fn: FakeUserResponseFn | None = None,
```
xingyaoww: Let's set the default value here directly

xingyaoww: @OpenHands please apply the change

benchmarks/utils/fake_user_response.py
```diff
@@ -0,0 +1,204 @@
...
+def _agent_sent_message(events: list[Event]) -> bool:
+    """Check if the agent's last event was a message (not a tool call).
+
+    Args:
+        events: List of conversation events.
+
+    Returns:
+        True if the last agent event was a MessageEvent, False otherwise.
+    """
+    for event in reversed(events):
```
xingyaoww, line 109: @OpenHands shall we just do reversed(events)[0] here? simiarlly in _agent_finished_with_finish_action

PR #311, opened 2026-01-14
Fix SDK submodule update to handle commit SHAs

juanmichelini:
This PR fixes the SDK submodule update step in the build-swebenchmultimodal-images workflow to properly handle commit SHAs.

**Problem:**
The workflow was trying to `git fetch origin <commit-sha>`, which doesn't work for commit SHAs - it only works for branch names and refs.

**Solution:**
- Fetch all from origin first to ensure the commit is available locally
- Then checkout the specific commit/ref directly
- This allows passing both branch names and commit SHAs to the sdk-commit input

**Testing:**
This fix is needed for the current evaluation run (20979458219) which is failing because the benchmarks build can't fetch the SDK commit.

### DIFF
--- a/benchmarks/swebenchmultimodal/build_images.py
+++ b/benchmarks/swebenchmultimodal/build_images.py
@@ -57,9 +57,17 @@ def extract_custom_tag(base_image: str) -> str:
     return name
 
 
-def collect_unique_base_images(dataset, split, n_limit):
+def collect_unique_base_images(
+    dataset,
+    split,
+    n_limit,
+    selected_instances_file: str | None = None,
+):
     df = get_dataset(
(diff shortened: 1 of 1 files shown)


juanmichelini:
@openhands to make this PR cleaner

keep the changes in

 .github/workflows/build-swebenchmultimodal-images.yml benchmarks/swebenchmultimodal/build_images.py 

but revert the changes in vendor/software-agent-sdk

neubig:
@openhands /codereview-roasted

neubig (changes requested):
Could we check if the above simplification that openhands suggested works?

juanmichelini:
Closing this as stale

PR #313, opened 2026-01-14
Add SWTBench profiling timelines

simonrosenberg:
## Summary
- add per-instance attempt timelines in evaluation orchestrator
- capture SWTBench eval/harness timings and emit structured JSON timelines
- inject profiling sitecustomize into swt-bench at runtime (optional via PROFILE_SWTBENCH)

## Testing
- not run (profiling-only instrumentation)

Note: vendor/software-agent-sdk submodule advanced to current main when updating base branch; no manual edits inside submodule.

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -21,7 +21,6 @@ from pydantic import BaseModel, Field
 from tqdm import tqdm
 
 from benchmarks.utils.constants import OUTPUT_FILENAME
-from benchmarks.utils.conversation import CONVERSATION_EVENT_LOGGING_ENV_VAR
 from benchmarks.utils.critics import get_completed_instances
 from benchmarks.utils.iterative import aggregate_results, get_failed_instances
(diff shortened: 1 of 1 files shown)


simonrosenberg:
@juanmichelini perhaps this could NOT be merged (because a bit messy) but used as a hack when we need to profile and optimize so branch... We can just ask OH to merge this into any other feature branch and then we get the profiling logs for free.. But I do enjoy the idea of adding precise profiling logs into the report.jsons. 
What is less good is that the profiling code depends on each benchmark and  Sitecustomize is hacky and messy so idk

juanmichelini:
> @juanmichelini perhaps this could NOT be merged (because a bit messy) but used as a hack when we need to profile and optimize so branch... We can just ask OH to merge this into any other feature branch and then we get the profiling logs for free.. But I do enjoy the idea of adding precise profiling logs into the report.jsons. What is less good is that the profiling code depends on each benchmark and Sitecustomize is hacky and messy so idk

Tough choice! Thoughs on doing a PR to the SWT bench repo?

Issue #277, opened 2026-01-07
Refine runtime failure classification for resource backoff


simonrosenberg:
Background
- PR #273 adds exponential resource_factor bumps after runtime failures. It currently relies on a hardcoded FATAL_RUNTIME_ERRORS string list.
- The SDK/runtime layer actually surfaces exceptions as httpx.HTTPStatusError (404/502/etc), ValueError for pod crashloop/failure, and RuntimeError for transient 'not yet ready'/health retries. No RuntimeUnavailable/RuntimeDisconnected exceptions are emitted.
- In large sweeps (e.g., issue #239 run eval-20705712290), failures were 404/502 and runtime not found/force-stops from runtime-api capacity/eviction. The string list misses those and over-matches generic RuntimeError.

Problem
- String matching on error text is brittle and misclassifies: it misses HTTP status-driven failures and can over-trigger on transient RuntimeError.
- We need a typed/status-based classifier to decide when to bump resource_factor versus just retry.

Proposal
- Introduce an exception classifier that treats:
  - fatal: httpx.HTTPStatusError with status in {404, 410, 500–504}; httpx.ConnectError/ReadTimeout/ConnectTimeout; ValueError from pod crashloop/failure; ConnectionError from sockets.
  - non-fatal: RuntimeError from 'not yet ready'/health retries.
- Ideally propagate structured signals from workspace (pod_status/status code) or raise typed exceptions so evaluation can decide cleanly.

Next steps
- Add classifier in evaluation utils and wire it into retry loop.
- Consider tightening SDK/workspace exceptions to carry status/pod_status for better classification.
- Revisit resource bump thresholds once classifier is in place.

PR #345, opened 2026-01-19
Add per-instance resource factor mapping for evaluation

xingyaoww:
## Summary

This PR adds support for per-instance resource factor mapping, porting over functionality from the [OpenHands main repository's SWE-bench evaluation](https://github.com/OpenHands/OpenHands/blob/e8ba741f28fd1683523ea2ed1e96d0c4c653dee4/evaluation/benchmarks/swe_bench/run_infer.py#L19).

## Motivation

Different benchmark instances may have varying resource requirements (e.g., large codebases, heavy dependencies). Some instances are known to be resource-intensive and consistently fail with default resource allocations, requiring multiple retries. By specifying per-instance base resource factors, we can:

1. **Avoid wasted retries**: Start resource-intensive instances with higher resources instead of failing with default resources
2. **Improve evaluation efficiency**: Reduce the number of retry attempts needed
3. **Better resource allocation**: Allocate resources based on actual instance needs

### Core Infrastructure
- **`benchmarks/utils/resource_mapping.py`**: New utility module to load per-instance resource factors from JSON files
  - `get_resource_mapping(benchmark_name, dataset_name)`: Loads and caches resource mappings
  - `get_instance_resource_factor(benchmark_name, dataset_name, instance_id, default_factor)`: Gets resource factor for a specific instance

### Evaluation Framework
- **`benchmarks/utils/evaluation.py`**:
  - Added `get_instance_base_resource_factor(instance)` method to `Evaluation` base class (can be overridden by subclasses)
  - Modified `_calculate_resource_factor()` to accept a `base_resource_factor` parameter
  - Updated retry logic to get per-instance base resource factor and use it for exponential backoff calculation
  - Enhanced logging to show both base and calculated resource factors

### SWE-bench Implementation
- **`benchmarks/swebench/run_infer.py`**: Implemented `get_instance_base_resource_factor()` override in `SWEBenchEvaluation` class
- **`benchmarks/swebench/resource/`**: New directory structure for resource mappings
  - `README.md`: Documentation on format and usage
  - `princeton-nlp__SWE-bench_Lite.json`: Example mapping with 4 resource-intensive instances

## How It Works

1. Each benchmark can have a `resource/` directory with JSON files mapping instance IDs to resource factors
2. File naming: dataset name with `/` replaced by `__` (e.g., `princeton-nlp__SWE-bench_Lite.json`)
3. JSON format: `{"instance-id": resource_factor}` where resource_factor is an integer ≥ 1
4. During evaluation:
   - Base resource factor is looked up per-instance (defaults to global `base_resource_factor` if no mapping exists)
   - On retry, resource factor is exponentially increased: `base_factor * 2^runtime_failure_count`
   - Capped at `max_resource_factor` from metadata (default: 8)

## Example

For an instance with base resource factor of 2:
- **First attempt**: resource_factor = 2
- **After 1st failure**: resource_factor = 4 (2 × 2¹)
- **After 2nd failure**: resource_factor = 8 (2 × 2²)
- **After 3rd failure**: capped at max_resource_factor (typically 8)

## Testing

Verified that:
- Resource mappings are correctly loaded from JSON files
- Per-instance resource factors are properly looked up
- Default fallback works for non-existent instances/datasets
- The implementation follows the same pattern as OpenHands main repository

## Backward Compatibility

✅ Fully backward compatible:
- Benchmarks without resource mappings continue to use the global `base_resource_factor`
- The default implementation of `get_instance_base_resource_factor()` returns the global value
- Only benchmarks that override the method and provide mapping files will use per-instance factors

## Future Work

- Add resource mappings for other benchmarks (commit0, multiswebench, etc.)
- Populate mappings for more datasets based on historical failure data
- Consider adding an exception classifier (as noted in TODO #277) to intelligently decide when to bump resources

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -29,7 +29,6 @@ from benchmarks.utils.models import (
     EvalOutput,
     RemoteRuntimeAllocation,
 )
-from benchmarks.utils.resource_mapping import get_instance_resource_factor
 from openhands.sdk import get_logger
 from openhands.sdk.critic import CriticBase
 from openhands.sdk.workspace import RemoteWorkspace
--- a/benchmarks/utils/resource_mapping.py
+++ b/benchmarks/utils/resource_mapping.py
@@ -6,10 +6,10 @@ This module provides utilities to track per-instance resource requirements.
 """
 
 import json
-import os
 from pathlib import Path
 
 from openhands.sdk import get_logger
+
 
 logger = get_logger(__name__)
 
@@ -56,9 +56,7 @@ def get_resource_mapping(
                 f"with {len(_global_resource_mapping[cache_key])} entries"
             )
         except Exception as e:
-            logger.warning(
-                f"Failed to load resource mapping from {mapping_file}: {e}"
-            )
+            logger.warning(f"Failed to load resource mapping from {mapping_file}: {e}")
             _global_resource_mapping[cache_key] = {}
             return None
 


xingyaoww:
@OpenHands please fix the failing actions on PR #345 at branch `add-per-instance-resource-factors`

xingyaoww:
hmm not sure if this is necessary now given that we already have the auto-increase on error

PR #362, opened 2026-01-26
Align Laminar runs with CI eval IDs

simonrosenberg:
## Summary
- set Laminar trace session/evaluation IDs from UNIQUE_EVAL_NAME (CI eval id)
- attach benchmark and model metadata to Laminar trace/eval payloads

## Testing
- uv run pre-commit run --files benchmarks/utils/laminar.py benchmarks/utils/evaluation.py

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -11,7 +11,7 @@ from concurrent.futures import ProcessPoolExecutor, as_completed
 from contextlib import contextmanager
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


simonrosenberg:
@openhands /codereview-roasted
Read this PR diff 
the goal is to set Laminar trace session/evaluation IDs from UNIQUE_EVAL_NAME (CI eval id) and attach benchmark and model metadata to Laminar trace/eval payloads

We should not add any useless complexity or useless code to the code
use pip to install laminar locally and inspect their code to understand how it works
and make sure this PR adds clean minimal code to reach the goal stated above

PR #389, opened 2026-02-02
Update run_id: remove "eval_" prefix 

simonrosenberg:
`run_id` had been modified previously to be directly equal to the github action run id without the `eval_` prefix.
Here there are fallbacks / bits of code that use the previous version of run_id.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2266,7 +2266,7 @@ wheels = [
 
 [[package]]
 name = "openhands-agent-server"
-version = "1.8.1"
+version = "1.10.0"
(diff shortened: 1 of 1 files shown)


juanmichelini (approved):
LGTM

simonrosenberg:
@openhands please resolve conflicts

simonrosenberg:
@openhands fix precommit hooks
Also make sure the PR has a clean diff: it should NOT how vendor changes. The diif should ONLY show removal of eval_

PR #394, opened 2026-02-04
fix(swebench-eval): allow disabling Modal via --no-modal

cizezsy:
The current implementation always set `modal` as true.  This PR adds the `--no-modal` option to swebench-eval cli to allow user to disable `modal`.

### DIFF
--- a/benchmarks/swebench/eval_infer.py
+++ b/benchmarks/swebench/eval_infer.py
@@ -246,15 +246,16 @@ Examples:
 
     parser.add_argument(
         "--modal",
+        dest="modal",
         action="store_true",
-        help="Use Modal for evaluation (default: True)",
+        help="Use Modal for evaluation",
     )
 
     parser.add_argument(
         "--no-modal",
-        action="store_false",
         dest="modal",
-        help="Disable Modal for evaluation",
+        action="store_false",
+        help="Do not use Modal for evaluation",
     )
 
     parser.add_argument(


juanmichelini:
@cizezsy thanks for your submission! I'm curious does --no-modal do the same as --modal false ?

cizezsy:
@juanmichelini Thanks for the reply. `--modal` uses `store_true`. So its value is `True`, If the flag is **present**, otherwise it uses default value specified in `EVAL_DEFAULTS`, which is also `True`. There is no way to change its value to `False`. Running the cli with `--modal false` throws errors.

```
uv run swebench-eval output.jsonl --run-id test --modal false

# Output
usage: swebench-eval [-h] [--dataset DATASET] [--output-file OUTPUT_FILE] [--skip-evaluation] [--workers WORKERS] [--split SPLIT] --run-id RUN_ID [--modal] [--timeout TIMEOUT] input_file
swebench-eval: error: unrecognized arguments: false
```

A simpler fix would be to change the default value of `modal` in `EVAL_DEFAULTS` to `False`. However, this would change the default behavior of the swebench-eval cli and could break backward compatibility. But, since the `--modal` flag was added only five days ago, the impact of this breaking change is low. I personally actually perfer to change the default value. What do you think?

juanmichelini:
@cizezsy thanks I see the problem now! @simonrosenberg will the current CI workflow work with this PR?
If so I propose we merge it so that @cizezsy can continue working. And maybe we can create an issue to standardise --no-modals or similar across benchmarks.

simonrosenberg:
> @cizezsy thanks I see the problem now! @simonrosenberg will the current CI workflow work with this PR? If so I propose we merge it so that @cizezsy can continue working. And maybe we can create an issue to standardise --no-modals or similar across benchmarks.

I dont think it makes sense to have both store_true/false AND a default value.
Perhaps we could 1. remove the defaults 2. update the logic so that EVAL_INFERs stores "store_true" and dynamically pass "store_true" instead of a default.

juanmichelini:
What about doing 

```
parser.add_argument(
    "--modal",
    dest="modal",
    action="store_true",
    help="Use Modal for evaluation",
)

parser.add_argument(
    "--no-modal",
    dest="modal",
    action="store_false",
    help="Do not use Modal for evaluation",
)

parser.set_defaults(modal=EVAL_DEFAULTS["modal"])
```

that way we can retain the CI workflow as is, retain defaults and be able to override them as needed?

simonrosenberg:
ah ok got it. I think we can simply remove `modal` and only keep `no-modal` then

juanmichelini:
@openhands please update this PR so that the default eval file is still read by defaut, but it can be overriden with --modal or --no-modal something like:
```

parser.add_argument(
    "--modal",
    dest="modal",
    action="store_true",
    help="Use Modal for evaluation",
)

parser.add_argument(
    "--no-modal",
    dest="modal",
    action="store_false",
    help="Do not use Modal for evaluation",
)

...

parser.set_defaults(modal=EVAL_DEFAULTS["modal"])
```

Please do minimal changes.

juanmichelini:
@openhands if neither --modal nor --no-modal is are used, it should be set to EVAL_DEFAULTS["modal"]
can you check this is so? and if not, update the PR?

simonrosenberg:
@openhands /codereview-roasted does the logic for modal/no-modal make sense? 
The arg parsing should
1. default to modal=True if nothing is passed
2. allow to set modal=False with --no-modal

juanmichelini:
Merged! Thanks @cizezsy for your help!

cizezsy:
Hi @juanmichelini , thanks for reviewing and merging the PR! I would like to clarify one small point regarding the removal of the default timeout value [here](https://github.com/OpenHands/benchmarks/pull/394/changes/23b60dbe208f80d999775673850f94071c88e64e). The default timeout are set at [eval_infer.py#L269](https://github.com/OpenHands/benchmarks/blob/main/benchmarks/swebench/eval_infer.py#L269). I think we should revert that commit so we don't have two places to set the default value. My apologies for not making this clearer in the PR description.

PR #409, opened 2026-02-11
fix(gaia): extract answers from FinishAction messages

neubig:
## Problem

The GAIA evaluation was reporting very low resolution rates (~23.6%) for some models like `jade-spark`, significantly lower than expected. After investigation, the root cause was identified as an **answer extraction failure**, not a model capability issue.

### Analysis

When analyzing the results from `litellm_proxy-jade-spark-2862` ([results](https://results.eval.all-hands.dev/gaia/litellm_proxy-jade-spark-2862/21885619897/results.tar.gz)):

- **Total instances**: 165
- **Instances with NO answer extracted**: 109 (66%)
- The evaluation log shows repeated failures:
  ```
  ERROR - Could not find agent output after 30 attempts and 1181.9s total wait time
  ```

The issue is that `_extract_answer_from_history` only looks for `MessageEvent` from the agent, but the agent frequently uses `FinishAction` directly with the solution in the message parameter, without first producing a separate `MessageEvent`.

Looking at actual conversation events:
- **Successful instances**: Have an agent `MessageEvent` containing `<solution>` tags
- **Failing instances**: Agent issues `FinishAction` with `<solution>142</solution>` in the message, but no separate `MessageEvent` - so the answer is never extracted

## Solution

Update `_extract_answer_from_history` to check both:
1. Agent `MessageEvent` with `llm_message` content (existing behavior, kept as priority)
2. `ActionEvent` with `FinishAction` containing a message (new fallback)

## Changes

- Added imports for `ActionEvent` and `FinishAction`
- Updated `_extract_answer_from_history` to also extract answers from `FinishAction.message`
- Maintained backward compatibility: `MessageEvent` is still preferred when present

## Expected Impact

This fix should significantly improve GAIA resolution rates for models that use `FinishAction` directly instead of producing a separate `MessageEvent`. The actual model capability is likely much higher than reported - possibly closer to the ~50%+ seen with other models on GAIA.

## Testing

- Verified Python syntax compiles correctly
- Logic follows the same pattern used in `fake_user_response.py` for detecting `FinishAction`

### DIFF
--- a/benchmarks/gaia/run_infer.py
+++ b/benchmarks/gaia/run_infer.py
@@ -493,33 +493,6 @@ For example: if you want to search for a research paper on Arxiv, either use the
                     )
                     return text
 
-            # Check for alternative output sources before retrying
-            if attempt == 0:
-                # Check for finish events that might contain output (legacy check)
-                finish_events = [
-                    e for e in events if "finish" in type(e).__name__.lower()
-                ]
-                if finish_events:
-                    logger.info(f"Found {len(finish_events)} finish events")
-                    for event in reversed(finish_events):
-                        output = getattr(event, "output", None)
-                        if output:
-                            logger.info(
-                                f"Found output in {type(event).__name__}: "
-                                f"{str(output)[:100]}"
-                            )
-                            return str(output)
-
-                # Check for error events
-                error_events = [
-                    e for e in events if "error" in type(e).__name__.lower()
-                ]
-                if error_events:
-                    logger.warning(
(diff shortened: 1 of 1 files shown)


neubig:
@openhands the checking logic has become very complex, could we simplify it to a single branch that catches both FinishAction and MessageAction?

juanmichelini:
testing it!

neubig:
We tested this and I think it looks good.

juanmichelini (approved):
LGTM

PR #410, opened 2026-02-11
fix: add timeout to evaluation loop to prevent indefinite stalls

neubig:
## Problem

Evaluation jobs were observed to hang indefinitely with no progress. Investigation of two stuck jobs revealed:

- `eval-21870785349-jade-spark`: Running 17h with 0/1 completions, 16 instances started but only 4 completed (all failures)
- `eval-21885622831-jade-spark`: Running 9h with 0/1 completions, 102 instances started but only 2 completed (all failures)

## Root Cause Analysis

**5 potential causes were investigated:**

1. **`fut.result()` without timeout** (CONFIRMED) - The main evaluation loop uses `as_completed(futures)` without a timeout parameter, meaning if any worker process hangs, crashes silently, or gets stuck in network operations, the main process will wait forever.

2. **Workspace cleanup blocking** - The `finally` block in `_process_one_mp` attempts to capture conversation archives and cleanup workspaces. If the runtime is in a bad state, these operations could hang.

3. **Runtime allocation waiting** - `prepare_workspace()` uses tenacity with a timeout for runtime allocation, so this is bounded.

4. **WebSocket polling without completion** - The `_wait_for_run_completion` method has a timeout (default 3600s), so this should eventually fail.

5. **`pool.shutdown(wait=True)` blocking** - After the tqdm loop, `pool.shutdown(wait=True)` waits for all workers indefinitely.

**Evidence from Datadog logs:**
- Many instances logged "Workspace does not support downloading directory" (archive capture step) then went silent
- The tqdm progress showed 0% completion even after hours
- No completion logs for most instances, indicating workers were stuck after evaluation but before returning

## Solution

Add a configurable timeout to the evaluation loop to ensure jobs complete within a bounded time:

1. Add `instance_timeout` field (default 4 hours) to the `Evaluation` class
2. Pass `timeout` parameter to `as_completed()` to limit how long we wait for futures
3. Track `future_to_instance` mapping to identify which instances timed out
4. Create proper error outputs for timed-out instances so they are recorded in results
5. Use `cancel_futures=True` in `pool.shutdown()` to clean up stuck workers

## Impact

- Evaluation jobs will now complete within a bounded time even if some workers hang
- Timed-out instances will be recorded as failures with proper error messages
- Results will be partial but usable instead of completely lost to a hung job

## Testing

This fix is targeted at production scenarios where workers hang due to:
- Network connectivity issues with runtime service
- Runtime pods getting evicted/killed
- Worker processes crashing silently (OOM, segfault)

The fix uses standard Python concurrent.futures timeout mechanisms and doesn't change the happy path behavior.

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -8,8 +8,9 @@ import os
 import sys
 import time
 from abc import ABC, abstractmethod
-from concurrent.futures import FIRST_COMPLETED, ProcessPoolExecutor, wait
+from concurrent.futures import FIRST_COMPLETED, Future, ProcessPoolExecutor, wait
 from contextlib import contextmanager
+from dataclasses import dataclass
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Callable, List, Optional, Tuple
@@ -38,6 +39,19 @@ from openhands.workspace import APIRemoteWorkspace
 
 logger = get_logger(__name__)
 
+# Interval in seconds between checking for per-instance timeouts
+TIMEOUT_CHECK_INTERVAL_SECONDS = 60
+
+
+@dataclass
+class PendingInstance:
+    """Tracks state for a pending evaluation instance."""
+
+    instance: EvalInstance
+    start_time: float
+    datapoint_id: UUID | None = None
+
+
 OnResult = Callable[[EvalInstance, EvalOutput], None]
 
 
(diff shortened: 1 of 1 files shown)


neubig:
@openhands /codereview-roasted /github-pr-review

neubig:
@openhands the review comments are good to some extent

juanmichelini (approved):
LGTM!

PR #412, opened 2026-02-12
(feat)NeMo-Evaluator Integration 

e-dobrowolska:
Changes introduced in the PR:

## 1. Packaging Improvements

* Modified `pyproject.toml` to add new CLI entrypoints, nemo-evaluator dependency and to include package files.
* Modified `benchmarks/utils/version.py` to gracefully handle SDK SHA extraction with fallback to `"unknown"` when git submodule information is unavailable. This prevents build failures in packaging scenarios where git metadata isn't present.
* Moved imports inside functions to avoid errors related to missing git metadata or package structure discoveries when the package is installed via pip without cloning the full repository.
* Added a fallback for loading files (especially `.jp2` images), so they can be found and loaded even when the package is installed with pip and the repo wasn’t cloned.

---

## 2. Unified Benchmark Execution

* Created `benchmarks/scripts/run_benchmark.py` as a single entrypoint to run both inference and evaluation for any benchmark in the repository.
* Created `benchmarks/scripts/generate_llm_config.py` script to create LLM config JSON files from command-line arguments.
* Made sure all benchmark scripts load LLM configs in the unified way, with an option to set API keys through environment variables.

---

## 3. Docker Image Handling

* Added `IMAGE_TAG_PREFIX` and `EVAL_AGENT_IMAGE` environment variable support to enable custom image naming and use of pre-built images from custom registries.
* All benchmark harnesses now have an option to check for image availability locally, and to use pre-built images.
* Created `benchmarks/utils/image_utils.py` with `image_exists()` to check for pre-built images locally.

---

## 4. New Evaluation Parameters

* Added `--conversation-timeout` parameter to all benchmark inference scripts to avoid timeout errros on long evaluation runs.
* Added `--skip-failed-samples` flag to control whether evaluation continues past individual sample failures or stops immediately on the first error.

---

## 5. UV Installation Handling

* If execution with `uv` fails due to it not being installed, the system will now gracefully retry using the standard Python interpreter.

---

## 6. Report format Standardization

* Standardized output report format across all benchmarks with consistent JSON structure including `benchmark_name`, `metadata`, `results`, and `metrics` fields.

---

## 7. Per-Benchmark Changes & Bugfixes

* **GAIA Critic Configuration**: Changed GAIA to use `"pass"` critic instead of `"finish_with_patch"` since it's a question-answering benchmark, not code generation.
* **OpenAgentSafety Pandas NaN Handling**: Fixed syntax/logic errors in `pd.isna()` checks to prevent crashes.
* **Error Raising**: Updated error handling to ensure all caught exceptions are now re-raised after logging, preventing silent failures and making issues visible during execution.
* **MultiSWEBench Docker Builds**: Fixed a bug in `benchmarks.multiswebench.build_images` where building images tried to load/prepare the HF dataset and crashed with `TypeError: Couldn't cast array of type ... to ...` (which then resulted in `datasets.exceptions.DatasetGenerationError: An error occurred while generating the dataset`).
* **MultiSWEBench Dataset Name Bug**: Replaced `dataset_path.startswith("ByteDance-Seed/Multi-SWE-bench"` with `"multi-swe-bench" in dataset_path.lower()` -- otherwise MultiSWEBench's `run_infer.py` would fail to detect that the `bytedance-research/Multi-SWE-Bench` dataset needs to be downloaded.

---

## 8. NeMo Evaluator Integration

NeMo Evaluator requires a specific directory structure and configuration files to recognize and run benchmark frameworks.

**What changed:**

* Created `nemo_evaluator/` directory with the following structure:
  * `nemo_evaluator/openhands_benchmarks/__init__.py`
  * `nemo_evaluator/openhands_benchmarks/framework.yml`
  * `nemo_evaluator/openhands_benchmarks/output.py`

**How it works now:**

* The NeMo Evaluator framework can now properly discover and load OpenHands benchmarks as a registered evaluation framework
* The `framework.yml` defines the benchmark configuration for NeMo Evaluator
* The `output.py` module handles output formatting compatible with NeMo Evaluator's reporting system

### DIFF
--- a/benchmarks/commit0/run_infer.py
+++ b/benchmarks/commit0/run_infer.py
@@ -25,13 +25,14 @@ from benchmarks.utils.evaluation_utils import (
     get_default_on_result_writer,
 )
 from benchmarks.utils.image_utils import image_exists
+from benchmarks.utils.llm_config import load_llm_config
 from benchmarks.utils.models import (
     EvalInstance,
     EvalMetadata,
     EvalOutput,
 )
 from benchmarks.utils.version import SDK_SHORT_SHA
-from openhands.sdk import LLM, Agent, Conversation, Tool, get_logger
+from openhands.sdk import Agent, Conversation, Tool, get_logger
 from openhands.sdk.workspace import RemoteWorkspace
 from openhands.tools.delegate import DelegateTool
 from openhands.tools.preset.default import get_default_tools
@@ -619,12 +620,7 @@ def main() -> None:
     if args.max_attempts < 1:
         raise ValueError(f"max_attempts must be >= 1, got {args.max_attempts}")
 
-    llm_config_path = args.llm_config_path
-    if not os.path.isfile(llm_config_path):
-        raise ValueError(f"LLM config file {llm_config_path} does not exist")
-    with open(llm_config_path, "r") as f:
-        llm_config = f.read()
-    llm = LLM.model_validate_json(llm_config)
+    llm = load_llm_config(args.llm_config_path)
     logger.info("Using LLM config: %s", llm.model_dump_json(indent=2))
 
     dataset_description = (
--- a/benchmarks/gaia/run_infer.py
+++ b/benchmarks/gaia/run_infer.py
@@ -28,10 +28,10 @@ from benchmarks.utils.evaluation_utils import (
 )
 from benchmarks.utils.fake_user_response import run_conversation_with_fake_user_response
 from benchmarks.utils.image_utils import image_exists
+from benchmarks.utils.llm_config import load_llm_config
 from benchmarks.utils.models import EvalInstance, EvalMetadata, EvalOutput
 from benchmarks.utils.version import SDK_SHORT_SHA
 from openhands.sdk import (
-    LLM,
     Agent,
     Conversation,
     Event,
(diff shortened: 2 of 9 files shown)


benchmarks/commit0/run_infer.py
```diff
@@ -185,28 +186,46 @@ def prepare_workspace(
...
+                    raise RuntimeError(
+                        f"On-the-fly build failed and pre-built image {agent_server_image} does not exist"
+                    )
+                
+                workspace = DockerWorkspace(
+                    server_image=agent_server_image,
+                    working_dir="/workspace",
+                    forward_env=forward_env or [],
+                )
+                logger.info(f"Using pre-built image {agent_server_image}")
```
simonrosenberg: Question: do we need this fallback? in which cases would the build fail? Could a retry mechanism not be enough?

benchmarks/commit0/run_infer.py
```diff
@@ -578,7 +597,10 @@ def evaluate_instance(
 
 def main() -> None:
     prompt_dir = (Path(__file__).parent / "prompts").resolve()
-    choices = [str(p.relative_to(Path.cwd())) for p in prompt_dir.glob("*.j2")]
+    try:
+        choices = [str(p.relative_to(Path.cwd())) for p in prompt_dir.glob("*.j2")]
+    except ValueError:
+        choices = [str(p) for p in prompt_dir.glob("*.j2")]
```
simonrosenberg: same question here about the defensive try/except. 
Why would this code fail? Does the logic need fixing, rather than adding a defensive fallback code path?

simonrosenberg:
Thanks for submitting this PR! 
It seems to be doing many different things. Can you break down this PR in a set of smaller independent PRs? 
Or perhaps: what is the main feature that this PR is adding? Could we reduce the scope of this PR to that core feature that you wish to add?

Issue #476, opened 2026-03-03
SWE-Bench image builds stuck for 5+ hours, blocking evaluation jobs


juanmichelini:
## Problem

Three GitHub Actions builds for SWE-Bench images have been stuck for **5+ hours** on the "Build and push SWE-Bench images" step, blocking evaluation jobs from running.

## Affected Builds

All three builds are for SDK commit `217b218272499aa5b17335d82627e090b67cf9aa` (Update orjson to 3.11.7):

| Run ID | Status | Elapsed (reported) | Actual Time Stuck |
|--------|--------|-------------------|-------------------|
| [22627544120](https://github.com/OpenHands/benchmarks/actions/runs/22627544120) | In Progress | 13s | 5+ hours |
| [22627606528](https://github.com/OpenHands/benchmarks/actions/runs/22627606528) | In Progress | 12s | 5+ hours |
| [22627616885](https://github.com/OpenHands/benchmarks/actions/runs/22627616885) | In Progress | 12s | 5+ hours |

## Impact

**Evaluation pods blocked:** Multiple evaluation pods have been waiting 5+ hours for the build to complete:
- `eval-22627492274-gemini-3-1-p9tlk` - swebench (5h22m waiting)
- `eval-22627517186-glm-5-k5xn4` - swebench (5h21m waiting)
- `eval-22627517186-qwen3-5-fl-n9vvw` - swebench (5h21m waiting)

Logs show continuous polling:
```
[2026-03-03 19:52:47 UTC] Benchmarks build run 22627544120: status=in_progress, conclusion=None
```

## Expected Behavior

Normal SWE-Bench image builds complete in **5-40 minutes**:
- Recent successful builds: 5m40s, 8m8s, 10m28s, 24m10s, 40m38s
- Longest recent build: 1h9m20s

## Build Details

**Workflow:** `build-swebench-images.yml`

**Stuck step:**
```yaml
- name: Build and push SWE-Bench images
  run: |
    uv run benchmarks/swebench/build_images.py \
      --dataset '${DATASET}' \
      --split '${SPLIT}' \
      --image ghcr.io/openhands/eval-agent-server \
      --push \
      --max-workers '${MAX_WORKERS}' \
      --max-retries '${MAX_RETRIES}'
```

**Job status shows:**
```
* Build and push SWE-Bench images (still running after 5 hours)
* Archive build logs (queued)
* Upload build logs (queued)
* Display build summary (queued)
```

## Possible Causes

[...]

PR #481, opened 2026-03-04
Optimize SWE-Bench image builds: add local Docker check before remote registry

juanmichelini:
## Summary

This PR adds a local Docker image check **before** the remote registry check in `build_image()`. This is a **performance optimization** that can reduce HTTP requests when building images that already exist locally.

## Investigation Update

After further investigation based on reviewer feedback:

1. **PR #471 only renamed functions** - it did not remove any local Docker check
2. **There was never a local check in `build_image()`** - the `image_exists()` function has always been a remote-only HTTP check
3. **The March 3rd stuck builds were transient** - builds before and after the stuck ones succeeded without any code changes
4. **Builds are now working** - the issue resolved without this PR being merged

## What This PR Actually Does

Adds an optimization to check local Docker images before making remote HTTP requests:

```python
# Check local Docker first (fast), then remote registry (slow)
if local_image_exists(t):
    logger.info("Image %s already exists locally. Skipping build.", t)
    return BuildOutput(base_image=base_image, tags=[t], error=None)
if remote_image_exists(t):
    logger.info("Image %s already exists in remote registry. Skipping build.", t)
    return BuildOutput(base_image=base_image, tags=[t], error=None)
```

**Benefits:**
- Skips HTTP requests for images that already exist locally
- Useful for local development and re-runs on the same machine
- Fast local Docker check (~100ms) vs slow HTTP request

**Limitations:**
- In CI environments with fresh runners, images typically don't exist locally, so the remote check will still be needed
- This is an optimization, not a fix for the transient stuck builds

## Recommendation

Given that:
1. The stuck builds were transient and resolved without this change
2. CI runners typically don't have pre-built images locally
3. The change adds complexity without fixing the root cause (which remains unknown)

**I recommend closing this PR.** If the team believes the local check optimization is valuable for local development workflows, we can keep it. Otherwise, this PR can be closed as the issue was transient.

## Testing

✅ All tests pass  
✅ Pre-commit checks pass

Related to #476 (but does not fix the root cause)

### DIFF
--- a/benchmarks/utils/build_utils.py
+++ b/benchmarks/utils/build_utils.py
@@ -303,16 +303,12 @@ def build_image(
         sdk_version=sdk_version,
     )
     for t in opts.all_tags:
-        # Check local Docker first (fast), then remote registry (slow)
-        # This avoids overwhelming the registry with 500+ concurrent HTTP requests
-        # when building large batches (e.g., all SWE-Bench images)
+        # Check local Docker first (fast ~100ms), then remote registry (slow HTTP)
         if local_image_exists(t):
             logger.info("Image %s already exists locally. Skipping build.", t)
             return BuildOutput(base_image=base_image, tags=[t], error=None)
         if remote_image_exists(t):
-            logger.info(
-                "Image %s already exists in remote registry. Skipping build.", t
-            )
+            logger.info("Image %s already exists in registry. Skipping build.", t)
             return BuildOutput(base_image=base_image, tags=[t], error=None)
     tags = build(opts)
     return BuildOutput(base_image=base_image, tags=tags, error=None)


simonrosenberg:
@openhands In the PR description it says 
"This commit renamed image_exists() to remote_image_exists() and removed the local Docker check. The build_image() function in build_utils.py now makes HTTP requests to GHCR for every image tag check, creating 500+ concurrent requests when building large batches."
However if you look at the PR itself https://github.com/OpenHands/benchmarks/pull/471
It does not do anything else then renaming functions. The commit removes something that was added in a previous commit in the same PR so it seems unlikely that this is the root cause of all our trouble.
Please keep investigating

juanmichelini:
@simonrosenberg I'm testing it here https://github.com/OpenHands/benchmarks/actions/runs/22685356917/job/65766732495

Images build normally at commit 744df225fe7a04775f5d755d07e9b5e363285628 with 500 images, but broke at some point after that. Maybe something not obvious related to skip_build? Notice that this happens with 500 images, and does not seem to happen with 50 images. So a lot of runs in between succeed.

juanmichelini:
@simonrosenberg this did not work, closing this PR. Will open a new one again when I have a tested fix.

Issue #531, opened 2026-03-18
SWT-Bench image building slowness: root cause analysis and fix plan


simonrosenberg:
## Root Cause

SWT-bench image build throughput regressed significantly. Late February runs (SDK `cefaebf`) built 364-382 images in 5h03m-5h29m (66-76 img/h). Mid-March runs built 314-409 images in 9h20m-9h57m (31-42 img/h). The root cause is **broken registry cache** due to a Dockerfile ARG ordering mistake.

### What happened

SDK PR [#2130](https://github.com/OpenHands/software-agent-sdk/pull/2130) (commit `fd80128`, Mar 3) added `OPENHANDS_BUILD_GIT_SHA` ARG **before** the `apt-get install` and `npm install` layers in `base-image-minimal`:

```dockerfile
FROM ${BASE_IMAGE} AS base-image-minimal
+ ARG OPENHANDS_BUILD_GIT_SHA=unknown   # changes every SDK bump
+ ENV OPENHANDS_BUILD_GIT_SHA=...
  RUN apt-get install ...                # cache key now includes SHA → miss
  RUN npm install ...                    # also busted (depends on apt-get)
```

### Why this matters

Benchmark images use **GHCR registry cache** (`--cache-from type=registry`). The cache tags are stable across SDK bumps (`buildcache-{target}-{base_image_slug}`, no SHA). Prior builds export layers with `--cache-to type=registry,mode=max`.

When the Dockerfile build graph matches, BuildKit reuses cached layers from GHCR — including the expensive `apt-get` and `npm install`. The ARG before these layers changes the ancestor chain hash, breaking layer matching even though the cache tag is found.

### Measured impact (from 414 build logs, run [#23164396524](https://github.com/OpenHands/benchmarks/actions/runs/23164396524))

- `apt-get install`: **103s mean** per image (rebuilt from scratch, 0/414 cached)
- `npm install`: **38s mean** per image
- Combined: **141s/image** of unnecessary rebuilds per SDK bump

### Secondary factor

SDK PR [#2465](https://github.com/OpenHands/software-agent-sdk/pull/2465) (commit `d129025`, Mar 16) added `npm install -g @zed-industries/claude-agent-acp @zed-industries/codex-acp` to every benchmark image. This is a new ~38s/image step that didn't exist in the Feb baseline. Not a bug per se, but compounds the cache invalidation problem.

## Fix

[...]

PR #515, opened 2026-03-13
fix: cache SDK sdist across image builds

simonrosenberg:
## Summary
- pre-build the vendored SDK sdist once per image-build run and pass it through to the SDK build helper
- keep the SDK-owned build path and telemetry intact by using the newer `prebuilt_sdist` API in the vendored SDK
- cover cached-sdist propagation in `build_image()` and `build_all_images()` tests

## Testing
- `uv run pre-commit run --files benchmarks/utils/build_utils.py tests/test_image_utils.py`
- `uv run pytest tests/test_image_utils.py`

Split out of #507.

PR #534, opened 2026-03-18
ci: increase SWT-Bench default parallelism to 16 workers / 50 batch

simonrosenberg:
## Summary

Raises the default parallelism for the SWT-Bench image build workflow:
- `max-workers`: 4 → 16
- `build-batch-size`: 15 → 50

These were reduced to 4/15 as a safety measure during the image build regression investigation. With the 24-hour timeout (#528) and sdist caching (#515) now merged, higher parallelism can be restored.

The inputs remain fully configurable — manual `workflow_dispatch` runs can still override to any value (e.g., `max-workers=4` for debugging).

**What this PR does NOT change:**
- No SDK submodule bump
- No cache mode changes (`OPENHANDS_BUILDKIT_CACHE_MODE` is not set — SDK defaults apply)

## Test plan

- [ ] Pre-commit checks pass
- [ ] Trigger a manual SWT-bench build to verify the new defaults work

Related: #531

### DIFF
--- a/.github/workflows/build-swtbench-images.yml
+++ b/.github/workflows/build-swtbench-images.yml
@@ -91,13 +91,9 @@ jobs:
       packages: write
       issues: write
 
-    # Defaults for automatic runs; keep INSTANCE_IDS/SELECT_FILE initialized so set -euo pipefail won't fail on unset vars.
+    # Initialize vars that may be empty so set -euo pipefail won't fail on unset refs.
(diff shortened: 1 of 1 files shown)


simonrosenberg:
Closing — the defaults are already configurable via workflow_dispatch inputs. Before changing defaults, we should do a proper grid search to find the optimal values.

Issue #540, opened 2026-03-19
Use cache-mode=off for batch benchmark image builds


simonrosenberg:
## Problem

Benchmark image builds spend significant time exporting BuildKit cache to the registry (`--cache-to`), but this cache is **never reused** because each SWE-bench instance has a unique base image and therefore a unique cache tag. The registry cache is 100% cold on every run (`cache_import_miss_count=1` for all images across all three experiment runs).

Meanwhile, Blacksmith's **sticky disk** already provides effective local layer caching — common Dockerfile steps (12-13 out of ~15) are cached locally across builds within the same run and across runs via Blacksmith's snapshot-and-clone mechanism.

## Experiment

Three SWT-bench builds (100 images each, `force-build=true`, `sdk-commit=main`) with different `OPENHANDS_BUILDKIT_CACHE_MODE` values:

| Cache mode | Build duration | Throughput | Avg cache export/image | Avg wall clock/image | Retries | Run |
|-----------|---------------|-----------|----------------------|---------------------|---------|-----|
| `max` | 2h 49m | 35.5 img/h | 22.4s (5.9%) | 380s | 1 | [#23275304313](https://github.com/OpenHands/benchmarks/actions/runs/23275304313) |
| `min` | 2h 37m | 38.2 img/h | 11.7s (3.4%) | 348s | 0 | [#23275495756](https://github.com/OpenHands/benchmarks/actions/runs/23275495756) |
| `off` | 1h 58m | 50.7 img/h | 0s (0%) | 263s | 0 | [#23275305744](https://github.com/OpenHands/benchmarks/actions/runs/23275305744) |

### Caveat: sticky disk inheritance makes `off` vs `max`/`min` not apples-to-apples

All three runs executed on **different Blacksmith physical hosts** but shared sticky disk state via Blacksmith's snapshot-and-clone mechanism. Sticky disk commit timestamps and disk sizes reveal the inheritance chain:

| Run order | Mode | Physical host | Sticky disk committed | Final disk used |
|-----------|------|--------------|----------------------|----------------|
| 1st | `max` | production-131.153.236.169 | 04:16:28 UTC | 355.10 GiB |
| 2nd | `min` | production-131.153.143.135 | 04:14:18 UTC | 360.24 GiB |
| 3rd | `off` | production-125.253.72.239 | 06:18:00 UTC | 546.29 GiB |

[...]

PR #543, opened 2026-03-19
ci(swtbench): change cache-mode default from off to max

### LINKED ISSUE #542: Dockerfile ARG ordering causes full rebuild on every SDK commit
## Problem

New SDK commits trigger 9+ hour full rebuilds because `OPENHANDS_BUILD_GIT_SHA` is declared **before** `apt-get`/`npm install` in `base-image-minimal`. BuildKit busts the cache on every layer after a changed ARG, so all 433 images re-run apt-get and npm even though those don't depend on the SDK.

## Fix

Move the ARG/ENV after the expensive layers: [SDK PR #2522](https://github.com/OpenHands/software-agent-sdk/pull/2522)

```diff
 FROM ${BASE_IMAGE} AS base-image-minimal
-ARG OPENHANDS_BUILD_GIT_SHA=unknown
-ENV OPENHANDS_BUILD_GIT_SHA=${OPENHANDS_BUILD_GIT_SHA}
 RUN apt-get update && apt-get install ...
 RUN npm install -g @zed-industries/...
+ARG OPENHANDS_BUILD_GIT_SHA=unknown
+ENV OPENHANDS_BUILD_GIT_SHA=${OPENHANDS_BUILD_GIT_SHA}
```

The existing `--cache-from` registry tags (`buildcache-{target}-{base_slug}`) don't include the SDK SHA, so base layers exported by a previous build are reusable across SDK commits.

## Test plan

1. Run SWT-bench build with `cache-mode=max` on SDK commit A (seeds registry cache)
2. Run SWT-bench build on SDK commit B with the Dockerfile fix — verify apt-get/npm are cache hits
3. Expected: ~30-45 min instead of 9+ hours

## Related

- Parent tracker: #531
- Supersedes the full base image split proposed in #538

simonrosenberg:
## Summary

Change SWT-bench `cache-mode` default from `off` to `max` so that base image layers are exported to the registry cache.

## Why

This is the counterpart to the Dockerfile ARG reorder fix (SDK PR [software-agent-sdk#2522](https://github.com/OpenHands/software-agent-sdk/pull/2522)). That fix makes `base-image-minimal` layers SDK-independent, so they can be reused across SDK commits via `--cache-from`. But `--cache-from` only works if a previous build exported those layers with `--cache-to`.

The prior rationale for `off` (#540) was that registry cache had a 100% miss rate. That was because `OPENHANDS_BUILD_GIT_SHA` busted the base layers on every SDK change. With the ARG reorder, base layers become cache-stable and registry export becomes useful.

## Test plan

- [ ] Run SWT-bench build with this PR + SDK #2522 on SDK commit A
- [ ] Run again on SDK commit B — verify base layers are cache hits and build time drops from ~9h to ~30-45m

Closes #542
Related: #541, #531

### DIFF
--- a/.github/workflows/build-swtbench-images.yml
+++ b/.github/workflows/build-swtbench-images.yml
@@ -64,9 +64,9 @@ on:
         default: '10'
         type: string
       cache-mode:
-        description: 'BuildKit cache export mode (max, min, off). Default: off'
+        description: 'BuildKit cache export mode (max, min, off). Default: max'
         required: false
-        default: 'off'
+        default: 'max'
         type: string
       force-build:
         description: 'Rebuild images even if matching remote tags already exist'


simonrosenberg:
Closed as duplicate of https://github.com/OpenHands/benchmarks/pull/547

PR #555, opened 2026-03-23
feat: three-phase image build with local docker for 4x speedup

simonrosenberg:
## Summary
Add a phased benchmark image build path that uses:

1. one shared builder image
2. per-instance base images
3. local final assembly with `docker build` + `docker push`

This path is enabled through the existing workflow input `use-prebuilt-bases` and is implemented with a dedicated entrypoint: `benchmarks/swebench/build_phased_images.py`.

## Current Behavior
When `use-prebuilt-bases=true`, the workflows now run the phased path:

- Phase 0: build and push `ghcr.io/openhands/eval-builder:{sdk_sha}`
- Phase 1: build and push `ghcr.io/openhands/eval-base:{instance_tag}`
- Phase 2: assemble final `ghcr.io/openhands/eval-agent-server:{sdk_sha}-{instance_tag}-source-minimal` images locally

Final assembly uses the thin `benchmarks/swebench/Dockerfile.agent-layer` Dockerfile and local Docker instead of remote `docker buildx --push`.

When `use-prebuilt-bases` is not enabled, the existing standard build path remains unchanged.

## What Is In This PR
- New phased-build entrypoint: `benchmarks/swebench/build_phased_images.py`
- Shared helper logic for:
  - building the shared builder image
  - building base images
  - assembling final images locally
- Workflow integration for both SWE-bench and SWT-bench
- Final-image wrapping still applied for repos that need the docutils/roman layer
- Updated SDK submodule to the SDK branch that exposes the required `builder` / `base-image-minimal` targets

## Why This Design
The heavy work is amortized into reusable images:
- the builder image changes with the SDK
- the base images change with the benchmark environments
- the final assembly step becomes lightweight and can reuse local Docker layer state across many images in one job

## Results
Validated workflow runs:

- Three-phase remote buildx assembly: run `23409630540`
  - about `455s/image`
- Three-phase local assembly: run `23409964288`
  - `87.12s/image` average for 100 images
- Three-phase local assembly: run `23417634029`
  - `70.654s/image` average, `61.4s` median for 500 images
- Three-phase local assembly: run `23417634506`
  - `69.019s/image` average, `59.2s` median for 433 images

## Dependency
Depends on:
- OpenHands/software-agent-sdk#2542

## Scope
This PR is about the phased local-assembly path only. The legacy direct-build path is still preserved as the default when the phased mode is not selected.

### DIFF
--- a/tests/test_phased_build.py
+++ b/tests/test_phased_build.py
@@ -11,9 +11,7 @@ from benchmarks.utils.build_utils import BuildOutput
 
 # ---------------------------------------------------------------------------
 # Helpers
-# Production uses ProcessPoolExecutor for true parallelism across builds.
-# Tests substitute ThreadPoolExecutor to avoid pickling issues with mocks.
-def _thread_pool(**kw):
+# ---------------------------------------------------------------------------
 
 
 def _ok_proc(stdout="", stderr=""):


simonrosenberg:
Addressed review feedback in 5592f9f:

**Fixed:**
- **#2 (Push error handling)**: Collect partial push failures instead of fail-fast. Reports pushed/total count and all error details.
- **#3 (Silent cleanup)**: Replaced `shutil.rmtree(ctx, ignore_errors=True)` with try/except that logs warnings.
- **#4 (Data structure)**: Replaced magic string `base_image="builder"` with `base_image="sdk-builder-stage"`.
- **#6 (Inconsistent conditionals)**: Standardized SWT-bench workflow to use `env.USE_PREBUILT_BASES == 'true'` matching SWE-bench pattern.
- Removed unused `platform` parameter from `assemble_agent_image`.

**Acknowledged, deferred:**
- **#1 (Testing gap)**: Valid concern. Build infrastructure integration tests need Docker and registry access which makes them hard to run in standard CI. Will create a follow-up issue for adding tests with a mock/stub Docker approach.
- **#5 (Workflow complexity)**: Agree the inline Python-in-YAML is fragile. Extracting to a `build_with_phases.py` script is the right call — will do in a follow-up to keep this PR focused on the optimization.

PR #564, opened 2026-03-24
fix: sanitize numpy types in _create_error_output to prevent silent result loss

### LINKED ISSUE #563: bug: _create_error_output silently drops all error results due to numpy serialization failure
## Bug

When an instance fails after exhausting all retries, `_create_error_output` includes `instance.data` (a `pandas` row converted via `row.to_dict()`) in the returned `EvalOutput`. Pandas row dicts contain numpy scalar types (`numpy.int64`, `numpy.float64`, `numpy.ndarray`, etc.). When pydantic tries to serialize this via `model_dump_json()`, it raises:

```
Failed to write to attempt file …/output.critic_attempt_1.jsonl:
  Unable to serialize unknown type: <class 'numpy.ndarray'>
on_result callback failed: Unable to serialize unknown type: <class 'numpy.ndarray'>
```

Both the file write **and** the `on_result` callback swallow this as a warning, so the result is **silently discarded**. The instance then appears as "never completed" in every subsequent critic attempt, gets retried endlessly, and accumulates zero output across all `n_critic_runs` passes.

Successful outputs produced by `evaluate_instance` do not include `instance.data`, so they serialize fine — only error outputs are affected.

## Observed impact

In a real 500-instance run of `swtbench` (`nvidia/nemotron-3-super-120b-a12b`), the model hit persistent HTTP 429 rate-limit errors. After retries were exhausted the numpy bug silently swallowed every error output:

| Metric | Value |
|---|---|
| Instances prepared | 433 |
| Instances with any output | **48** |
| Instances silently lost | **385** |
| `Failed to write to attempt file` warnings | 1,220 |
| `on_result callback failed` warnings | 1,220 |
| Total inference wall time | **28.65 hours** |

## Root cause

`benchmarks/utils/evaluation.py`, `_create_error_output`:

```python
def _create_error_output(self, instance, error, retry_count):
    return EvalOutput(
        instance_id=instance.id,
        test_result={},
        instruction=None,
        error=f"Instance failed after {retry_count} retries. Last error: {str(error)}"[:200],
        history=[],
        instance=instance.data,   # ← pandas row dict — contains numpy types
    )
```

## Fix

Sanitize `instance.data` through a recursive converter that maps numpy scalars / arrays to their Python equivalents before storing them in `EvalOutput`.

juanmichelini:
## Problem

Fixes #563

When an instance fails after exhausting all retries, `_create_error_output` stores `instance.data` (a `pandas` `row.to_dict()`) in `EvalOutput`. Pandas row dicts contain numpy scalar types (`np.int64`, `np.float64`, `np.ndarray`, etc.) that Pydantic's `model_dump_json` cannot serialise:

```
WARNING  Failed to write to attempt file …/output.critic_attempt_1.jsonl:
         Unable to serialize unknown type: <class 'numpy.ndarray'>
WARNING  on_result callback failed: Unable to serialize unknown type:
         <class 'numpy.ndarray'>
```

Both the attempt-file write **and** the `on_result` callback swallow this as a warning, so the result is **silently discarded**. The instance then appears as "never completed" in every subsequent critic attempt, gets retried across all `n_critic_runs` passes, and still produces zero output at the end.

### Observed impact (real run)

A 500-instance `swtbench` run with `nvidia/nemotron-3-super-120b-a12b` hit persistent HTTP 429 rate-limit errors. After retries were exhausted the numpy bug silently swallowed every error output:

| Metric | Value |
|---|---|
| Instances prepared | 433 |
| Instances with **any** output | **48** |
| Instances silently lost | **385** |
| `Failed to write to attempt file` warnings | 1,220 |
| `on_result callback failed` warnings | 1,220 |
| Total inference wall time | **28.65 hours** |

Successful outputs (from `evaluate_instance`) do not include `instance.data`, so they serialised fine — only error outputs were affected.

## Fix

Add a `_to_serializable()` helper that recursively converts numpy types to their Python equivalents, and apply it to `instance.data` in `_create_error_output`:

```python
def _to_serializable(obj: Any) -> Any:
    if isinstance(obj, np.integer):  return int(obj)
    if isinstance(obj, np.floating): return float(obj)
    if isinstance(obj, np.bool_):    return bool(obj)
    if isinstance(obj, np.ndarray):  return obj.tolist()
    if isinstance(obj, dict):        return {k: _to_serializable(v) for k, v in obj.items()}
    if isinstance(obj, (list, tuple)): return type(obj)(_to_serializable(v) for v in obj)
    return obj
```

## Tests

`tests/test_error_output_serialization.py` covers:
- Each numpy type (`int64`, `float64`, `bool_`, `ndarray`)
- Nested dicts and list/tuple containers
- Plain Python passthrough (no-op)
- Integration: `_create_error_output` with a numpy-laden `instance.data` round-trips cleanly through `model_dump_json` / `json.loads`

### DIFF
--- a/tests/test_error_output_serialization.py
+++ b/tests/test_error_output_serialization.py
@@ -79,10 +79,10 @@ def test_to_serializable_list_and_tuple():
 
 def _make_eval(tmp_path):
     """Instantiate a minimal concrete Evaluation subclass for testing."""
-    from unittest.mock import MagicMock
-
     from benchmarks.utils.evaluation import Evaluation
     from benchmarks.utils.models import EvalMetadata
+    from openhands.sdk.critic import PassCritic
+    from openhands.sdk.llm import LLM
 
     class _DummyEval(Evaluation):
         def prepare_instances(self):
@@ -94,17 +94,9 @@ def _make_eval(tmp_path):
         def evaluate_instance(self, instance, workspace):
             raise NotImplementedError
 
-    llm_mock = MagicMock()
-    llm_mock.model = "test-model"
-    llm_mock.model_dump_json.return_value = "{}"
-    llm_mock.model_copy.return_value = llm_mock
-    llm_mock.temperature = 0.0
-
-    critic_mock = MagicMock()
-    critic_mock.model_dump.return_value = {"kind": "mock"}
-
+    llm = LLM(model="test-model")
     metadata = EvalMetadata(
-        llm=llm_mock,
+        llm=llm,
         dataset="test_ds",
         dataset_split="test",
         max_iterations=10,
@@ -112,7 +104,7 @@ def _make_eval(tmp_path):
         details={},
         eval_limit=1,
         max_retries=1,
-        critic=critic_mock,
+        critic=PassCritic(),
     )
     return _DummyEval(metadata=metadata, num_workers=1)
 


juanmichelini:
Testing here https://openhands-eval-monitor.vercel.app/?run=swebench%2Flitellm_proxy-nvidia-nemotron-3-super-120b-a12b%2F23469238479%2F&text=fix

juanmichelini:
Tested.

juanmichelini:
@openhands please fix pre commit checks

PR #569, opened 2026-03-25
fix: remove redundant 'Apply workflow_dispatch overrides' step from build workflows

### LINKED ISSUE #565: Remove redundant 'Apply workflow_dispatch overrides' step from build workflows
## Problem

Several build workflows have a redundant pattern: an `env:` block with hardcoded defaults, followed by an "Apply workflow_dispatch overrides" step that copies `inputs.*` values into `GITHUB_ENV`. This override step is effectively a no-op because:

1. For `workflow_dispatch`: inputs always have defaults that match the env block
2. For `pull_request_target`: the step is skipped entirely (gated by `if: workflow_dispatch`)

The duplication is error-prone — if someone updates the input default without updating the env default (or vice versa), they silently diverge.

## Fix

Replace hardcoded env defaults with `${{ inputs.X || 'default' }}` expressions. For `pull_request_target`, inputs are empty so the `||` fallback applies. For `workflow_dispatch`, the input value is used directly. This eliminates the need for the separate override step.

This was already done for `build-swtbench-images.yml` in PR #555. The same cleanup should be applied to:

- [ ] `.github/workflows/build-swebench-images.yml`
- [ ] Any other workflows using this pattern

## Example

Before:
```yaml
env:
  DATASET: 'princeton-nlp/SWE-bench_Verified'
# ...
steps:
  - name: Apply workflow_dispatch overrides (if any)
    if: ${{ github.event_name == 'workflow_dispatch' }}
    run: |
      if [ -n "${{ inputs.dataset }}" ]; then echo "DATASET=${{ inputs.dataset }}" >> "$GITHUB_ENV"; fi
```

After:
```yaml
env:
  DATASET: ${{ inputs.dataset || 'princeton-nlp/SWE-bench_Verified' }}
```

ixchio:
## Summary

This PR removes the redundant 'Apply workflow_dispatch overrides' step from all build workflows and replaces it with direct input references using the `${{ inputs.X || 'default' }}` pattern.

## Problem

Several build workflows had a redundant pattern: an `env:` block with hardcoded defaults, followed by an "Apply workflow_dispatch overrides" step that copies `inputs.*` values into `GITHUB_ENV`. This override step is effectively a no-op because:

1. For `workflow_dispatch`: inputs always have defaults that match the env block
2. For `pull_request_target`: the step is skipped entirely (gated by `if: workflow_dispatch`)

The duplication is error-prone — if someone updates the input default without updating the env default (or vice versa), they silently diverge.

## Solution

Replace hardcoded env defaults with `${{ inputs.X || 'default' }}` expressions:
- For `workflow_dispatch`: the input value is used directly
- For `pull_request_target`: inputs are empty, so the `||` fallback applies

This was already done for `build-swtbench-images.yml` in PR #555. This PR applies the same cleanup to:

- [x] `build-swebench-images.yml`
- [x] `build-commit0-images.yml`
- [x] `build-multiswebench-images.yml`
- [x] `build-swebenchmultimodal-images.yml`
- [x] `build-swegym-images.yml`
- [x] `build-swesmith-images.yml`

## Changes

1. **Removed** the "Apply workflow_dispatch overrides" step from all 6 workflows
2. **Updated** "Build selected instances file" steps to use `inputs.instance-ids` directly
3. **Updated** "Build and push" steps to define variables inline with fallback defaults
4. **Preserved** label-based N_LIMIT logic for `pull_request_target` events (for workflows that have it)

## Testing

The workflows should behave identically to before:
- For `workflow_dispatch`: inputs are used directly with their defaults
- For `pull_request_target`: fallback defaults are used (same as the previous env block defaults)
- Label-based N_LIMIT adjustments continue to work via GITHUB_ENV

Fixes #565

### DIFF
--- a/.github/workflows/build-commit0-images.yml
+++ b/.github/workflows/build-commit0-images.yml
@@ -104,21 +104,11 @@ jobs:
           ref: ${{ steps.checkout-ref.outputs.ref }}
           submodules: recursive
 
-      - name: Apply workflow_dispatch overrides (if any)
-        if: ${{ github.event_name == 'workflow_dispatch' }}
-        run: |
-          if [ -n "${{ inputs.dataset }}" ]; then echo "DATASET=${{ inputs.dataset }}" >> "$GITHUB_ENV"; fi
-          if [ -n "${{ inputs.split }}" ]; then echo "SPLIT=${{ inputs.split }}" >> "$GITHUB_ENV"; fi
-          if [ -n "${{ inputs.repo-split }}" ]; then echo "REPO_SPLIT=${{ inputs.repo-split }}" >> "$GITHUB_ENV"; fi
-          if [ -n "${{ inputs.max-workers }}" ]; then echo "MAX_WORKERS=${{ inputs.max-workers }}" >> "$GITHUB_ENV"; fi
-          if [ -n "${{ inputs.n-limit }}" ]; then echo "N_LIMIT=${{ inputs.n-limit }}" >> "$GITHUB_ENV"; else echo "N_LIMIT=" >> "$GITHUB_ENV"; fi
-          if [ -n "${{ inputs.instance-ids }}" ]; then echo "INSTANCE_IDS=${{ inputs.instance-ids }}" >> "$GITHUB_ENV"; fi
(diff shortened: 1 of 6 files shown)


simonrosenberg:
Hi @ixchio !
Thanks a lot for your fix. Apologies but I decided to include that change in this PR https://github.com/OpenHands/benchmarks/pull/555 so I am closing this one... 
Thanks again and have a nice day :)

PR #594, opened 2026-03-30
feat(acp): add acp-gemini agent type for Gemini CLI

simonrosenberg:
## Summary

Adds `acp-gemini` as a new agent type for running Gemini CLI as an ACP agent in benchmarks.

### ACP utilities (`benchmarks/utils/acp.py`)
- Command: `["gemini", "--acp"]`
- Env vars forwarded: `GEMINI_API_KEY`, `GEMINI_BASE_URL`
- `add_acp_agent_metadata()`: falls back to `conversation.state.agent_state` when local agent PrivateAttrs are empty (remote conversations)

### CLI and types
- `agent_type` Literal: added `"acp-gemini"`
- `--agent-type` CLI choices: added `"acp-gemini"`

### Docker
- `Dockerfile.agent-layer`: installs ACP servers (`claude-agent-acp`, `codex-acp`, `@google/gemini-cli`) in Phase 2 assembly to bypass cached base images that may lack newly added servers

### All 5 benchmark runners updated
- `swebench`, `gaia`, `swtbench`, `commit0`, `swebenchmultimodal`: pass `agent_state` to `add_acp_agent_metadata()`

### Tests
- 4 new tests for acp-gemini (25/25 total pass)

## Test plan
- [x] 25/25 benchmark unit tests pass (including 4 new gemini tests)
- [x] Pre-commit hooks pass (ruff, pyright, pycodestyle)
- [x] SWE-bench + GAIA benchmarks validated with acp-gemini

### DIFF
--- a/benchmarks/utils/Dockerfile.agent-layer
+++ b/benchmarks/utils/Dockerfile.agent-layer
@@ -35,12 +35,5 @@ COPY --from=builder /usr/local/lib/libpython3.13.so* /usr/local/lib/
 COPY --from=builder /usr/local/bin/python3 /usr/local/bin/python3
 ENV LD_LIBRARY_PATH=/usr/local/lib
 
-# Ensure ACP servers are present even when the eval-base image was cached
-# before a new server was added (eval-base tags don't include SDK SHA).
-# npm install -g is idempotent — no-op for already-installed packages.
-USER root
-RUN npm install -g @zed-industries/claude-agent-acp @zed-industries/codex-acp @google/gemini-cli
-USER ${USERNAME}
-
 COPY --chown=${USERNAME}:${USERNAME} --from=builder /agent-server /agent-server
 ENTRYPOINT ["/agent-server/.venv/bin/python", "-m", "openhands.agent_server"]


benchmarks/utils/acp.py
```diff
@@ -118,10 +120,29 @@ def build_acp_agent(agent_type: str, llm_model: str) -> ACPAgent:
...
+) -> None:
+    """Add ACP agent metadata to an eval result payload.
+
+    The ACPAgent stores agent_name/version in ``state.agent_state`` during init.
+    This syncs back in the ``full_state`` ConversationStateUpdateEvent.  We scan
+    the conversation events (last to first) to extract it.
+    """
+    name = ""
+    version = ""
+    for ev in reversed(list(conversation.state.events)):
```
enyst, line 135: I understand we don’t have in benchmarks really long running conversations, just to note, if we ever can benchmark those, `list(events)` is O(n) and it will cry.

Please don’t take it as a review comment, just a note for the future maybe

benchmarks/utils/acp.py
```diff
@@ -118,10 +120,29 @@ def build_acp_agent(agent_type: str, llm_model: str) -> ACPAgent:
...
-    test_result["acp_agent_version"] = cast(Any, agent).agent_version
+def add_acp_agent_metadata(
+    test_result: dict[str, Any],
+    conversation: Any,
+) -> None:
+    """Add ACP agent metadata to an eval result payload.
+
+    The ACPAgent stores agent_name/version in ``state.agent_state`` during init.
+    This syncs back in the ``full_state`` ConversationStateUpdateEvent.  We scan
+    the conversation events (last to first) to extract it.
```
enyst, line 131: I wonder if we have or should have, some easier and standard way to get the first and the last `ConversationStateUpdateEvent` - this is not the first need for it that I see… 🤔

I think maybe it’s worth looking into some small API to search for an event of a type, if we don’t already have one in the SDK 🤔. I think we have a `/search` in the agent-server interface, maybe python API is a bit behind

@openhands clone yourself the agent-sdk from the same org and investigate the two APIs, Python and REST, that we offer for working with conversations. Analyze the python API from a client developer perspective, and file an issue with your analysis and proposals. Do not modify this PR, just file the issue on the SDK repo.

Issue #617, opened 2026-04-03
async eval worker count is effectively capped below num_workers by default asyncio threadpool


simonrosenberg:
## Summary

The async evaluation path in `benchmarks` is configured with `num_workers=30` for benchmarks like `gaia`, `swebench`, and `swtbench`, but the live eval jobs are not actually achieving anything close to 30-way instance concurrency.

The main issue is that the async runner uses `asyncio.to_thread(...)` without installing a larger executor, so effective concurrency is bounded by Python's default `ThreadPoolExecutor` size rather than by `num_workers`.

## Relevant code

- `benchmarks/utils/evaluation.py`
  - `asyncio.Semaphore(self.num_workers)`
  - `await asyncio.to_thread(self._process_one_sync, ...)`
- benchmark defaults
  - `benchmarks/gaia/config.py` -> `num_workers = 30`
  - `benchmarks/swebench/config.py` -> `num_workers = 30`
  - `benchmarks/swtbench/config.py` -> `num_workers = 30`

## Live evidence

I verified this directly on the currently running eval jobs for ACP Gemini:

- `gaia` job `eval-23925785249-gemini-3-f`
- `swebench` job `eval-23925785262-gemini-3-f`

Inside both live eval pods:

```bash
python3 -c 'import os,concurrent.futures as f; ex=f.ThreadPoolExecutor(); print(os.cpu_count()); print(len(os.sched_getaffinity(0))); print(ex._max_workers); ex.shutdown()'
```

Output in both pods:

```text
4
4
8
```

Also from cgroup quota in both pods:

```text
cpu.max = 400000 100000
```

So the eval pod is capped at 4 CPUs, and Python's default threadpool size is only 8.

That means:

- requested concurrency: `30`
- actual `to_thread` worker ceiling: `8`

## Why this matters

`Evaluation._run_attempt_async()` schedules one task per instance, but each task immediately enters `asyncio.to_thread(self._process_one_sync, ...)`. Since `_process_one_sync()` then runs the full instance synchronously, each occupied executor thread is held for the duration of the instance.

For ACP runs this is especially expensive because each instance effectively blocks on a long `RemoteConversation.run()` / ACP prompt loop.

So the current architecture makes `num_workers=30` misleading in practice unless the executor is explicitly sized to match it.

## Additional live observations

The live jobs also show runtime retries, which further reduce throughput:

- `gaia` had many `runtime init failure` / `Remote conversation ended with error` / `ACPPromptError: terminated` messages
[...]

PR #618, opened 2026-04-03
Respect configured infer worker capacity

simonrosenberg:
## Summary
- install an explicit asyncio thread executor sized from `num_workers`
- log configured worker count, effective executor capacity, visible CPU count, and `cpu.max`
- add focused tests for executor sizing and startup telemetry

## Why
Live eval jobs were configured with `workers=30` but were only reaching about `7-8` concurrent infer runs while they still had many instances left, which rules out end-of-job drain:
- `eval-23925785249-gemini-3-f-n5rz8` (GAIA): `8/30`, `86` remaining
- `eval-23925785262-gemini-3-f-hgj8w` (SWE-bench): `7-8/30`, `353` remaining

That matches Python's default `ThreadPoolExecutor` sizing on a 4-vCPU pod (`min(32, cpu+4) = 8`). Since the async evaluation path relies on `asyncio.to_thread(...)`, effective inference concurrency was being capped by the implicit default executor rather than by the configured benchmark worker count.

## Testing
- `uv run pytest /Users/simonrosenberg/benchmarks/tests/test_async_evaluation.py`
- `uv run ruff check /Users/simonrosenberg/benchmarks/benchmarks/utils/evaluation.py /Users/simonrosenberg/benchmarks/tests/test_async_evaluation.py`

## Related
- OpenHands/evaluation#406
- OpenHands/benchmarks#617

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -15,6 +15,7 @@ import os
 import threading
 import time
 from abc import ABC, abstractmethod
+from concurrent.futures import ThreadPoolExecutor
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Superseded by clean main-based PR #619.

PR #672, opened 2026-04-19
fix(swtbench): rmi after push + free runner disk (fixes evaluation#495)

simonrosenberg:
## Summary

Fixes OpenHands/evaluation#495. SWT-bench image builds started hitting `No space left on device` on `ubuntu-latest-8core` partway through assembly, e.g. [run 24572967778](https://github.com/OpenHands/evaluation/actions/runs/24572967778/job/71850799413) (disk filled at image 160/433 after ~42 min).

Root cause is a latent bug in `assemble_agent_image` exposed by the Blacksmith → GitHub-hosted migration in #651:

- Assembly uses `docker build` + `docker push` (comment documents the 455 s → 70 s speedup vs. `buildx --push`) but never calls `docker rmi` on the assembled tag. Every per-instance agent image stayed in `/var/lib/docker`.
- Blacksmith runners had hundreds of GB of persistent disk, so accumulation never mattered. `ubuntu-latest-8core` has ~80 GB and fills up around image 150.
- Cache-hit and `n-limit=5` runs hid this because `remote_image_exists(final_tag)` short-circuits the docker build. No full cold SWT-bench run has actually succeeded on `ubuntu-latest-8core` since the migration.

## Changes

1. **`benchmarks/swebench/build_base_images.py`** — in `assemble_agent_image`, after all pushes succeed, run `docker rmi -f <pushed tags>`. Shared base+builder layers stay cached (they're what the comment-advertised speedup relies on), only the just-pushed final tag is dropped. Failures warn but don't fail the assembly.
2. **`.github/workflows/build-swtbench-images.yml`** — belt-and-suspenders headroom so the next unexpected growth doesn't OOM again:
   - `jlumbroso/free-disk-space` pinned to `54081f138730dfa15788a46383842cd2f914a1be` (v1.3.1). Frees ~30 GB by removing preinstalled Android/.NET/Haskell/etc. Keeps the docker-images cache.
   - Swebench-style preflight step that prunes BuildKit cache to 60 GiB target and fails fast if less than 75 GB free on `/var/lib/buildkit`.

The code change is the actual fix; the workflow changes are defense in depth (applying them to SWE-bench too should follow in a separate PR since SWE-bench has the same latent bug).

## Testing

- Existing `TestAssembleAgentImage` tests should still pass — mocked `subprocess.run` returns OK for the new `rmi` call; assertions only check `result.tags` / `result.error`.
- Real validation: re-running [evaluation run 24572967778](https://github.com/OpenHands/evaluation/actions/runs/24572967778) with this branch pointed at from the evaluation repo. Will comment back on the issue with the re-run URL.

## AI disclosure

This PR was prepared by Claude (Claude Code) on behalf of @simonrosenberg.

### DIFF
--- a/.github/workflows/build-swtbench-images.yml
+++ b/.github/workflows/build-swtbench-images.yml
@@ -205,37 +205,27 @@ jobs:
         run: |
           make build
 
-      - name: "Preflight: prune cache and verify BuildKit disk"
+      # Runs right after setup-buildx creates the docker-container builder, so
+      # the buildx cache is empty — this step is just a sanity gate on
+      # root-fs free space. Earlier revisions also ran `docker buildx prune
+      # --max-storage` here (unsupported flag, always failed over to
+      # --keep-storage and reclaimed 0 B) and a `df /var/lib/buildkit` check
+      # (path doesn't exist on GH-hosted runners because BuildKit runs in the
+      # buildx container, so the check always printed "skipping"). Both were
+      # removed.
+      - name: "Preflight: verify runner disk"
         run: |
           set -euo pipefail
-          KEEP_GB=60
-          echo "Pruning BuildKit cache (target max-storage ${KEEP_GB} GiB, no filters)..."
-          if ! docker buildx prune --all --force --max-storage ${KEEP_GB}g; then
-            docker buildx prune --all --force --keep-storage ${KEEP_GB}g || true
-          fi
-
           df -h / || true
           docker system df || true
-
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Superseded by #690, which keeps the disk fix in the shared SWE/SWT phased image assembly code without workflow preflight changes.

simonrosenberg:
Closing in favor of #690 so the SWE-bench and SWT-bench disk cleanup fix is reviewed as one shared-code PR.

Issue #684, opened 2026-04-21
GHCR `eval-agent-server` has 133k+ tags with no retention; tag scheme fans out per SDK sha


VascoSch92:
## Problem

The package `ghcr.io/openhands/eval-agent-server` currently holds **133,445 tags** with no retention policy, accumulated over ~5 months. The tag scheme `{sdk_sha}-…-{bench}-{repo}-source-minimal` fans out on every SDK commit, so each new sha triggers a full rebuild+push of the whole benchmark dataset. This is both an ergonomic problem (GHCR UI and tag APIs are unusable at this scale) and a concrete pipeline problem (builds now hit ENOSPC on GH-hosted runners).

The package is public, so this is **not** a billing/quota issue — but it is actively hurting the build pipeline.

## Evidence (as of 2026-04-21)

| Metric | Value |
|---|---|
| Total tags | 133,445 |
| SDK-sha image tags | 130,602 (97.9%) |
| `buildcache-*` tags | 2,793 |
| Legacy `v1.0.0*` tags | 30 |
| Distinct SDK shas producing tags | 548 |
| SDK shas with ≥ 1,000 tags each | 63 |
| Oldest SDK sha still present | `db900ea` — committed 2025-11-24 |
| Median per-tag compressed size | 1.38 GB (multi-arch index) |
| Buildcache median size | 2.03 GB |

Benchmark distribution (substring match):
- `swebench`: 130,304 tags (97.6%)
- `commit0`: 1,435
- `gaia`: 192

## Downstream effect

This directly causes the build hangs we've been seeing. Example: eval run `OpenHands/evaluation` Actions run [24706920746](https://github.com/OpenHands/evaluation/actions/runs/24706920746) (commit0 × Gemini 3.1 Pro) has been stuck in `Build Commit0 Images / build-and-push` for 4h45m+ because its SDK sha `aabf4072…` has no cached tags in the registry → BuildKit has to build + push the full image set → GH `ubuntu-24.04` runner (~14 GB free) hits ENOSPC → BuildKit hangs instead of failing fast.

The prior failed run on the same branch (`24682483119`) has this in its `manifest.jsonl` artifact:

```
"built", "docker.io/wentingzhao/minitorch:v0", "error": "OSError(28, 'No space left on device')"
```

## Reproduce the numbers

Anonymous (no GH token needed; package is public):

```bash
[...]

PR #689, opened 2026-04-22
Add nightly GHCR retention workflow for eval-agent-server

VascoSch92:
## Summary

Adds `.github/workflows/ghcr-retention.yml` — a nightly GC job that deletes old versions of `ghcr.io/openhands/eval-agent-server`. Refs #684.

## What it does

- Nightly at 03:17 UTC, plus `workflow_dispatch` with `dry-run` (default `true`), `cut-off` (default `2mo`), and `keep-n-most-recent` (default `100`) inputs. Scheduled runs perform real deletions; manual runs dry-run by default.
- Deletes versions older than `cut-off` **except** the 100 most-recently-created and any tag matching `v1.0.0*`.
- Uses [`snok/container-retention-policy@v3.0.0`](https://github.com/snok/container-retention-policy) — handles multi-arch manifest lists so deleting a parent index doesn't leave dangling children.
- Scoped to `eval-agent-server` only.

## Prerequisites

1. Add secret `GHCR_CLEANUP_PAT` — classic PAT with `delete:packages` + `read:packages` (or fine-grained with `Packages: read & write` on the org). `GITHUB_TOKEN` cannot delete org package versions.
2. Confirm `benchmarks` has `Write` role at https://github.com/orgs/OpenHands/packages/container/eval-agent-server/settings (the existing 8 build workflows already push successfully, so this is likely already set).

## Rollout

1. Merge.
2. Trigger manually with `dry-run: true` → read the log, sanity-check the delete count (expect tens of thousands).
3. Trigger manually with `dry-run: false` for the first real purge, or wait for the next nightly. First run may take hours due to GHCR secondary rate limits (`timeout-minutes: 720`).

## Known gap

Retention does not cross-reference eval metadata on GCS (`gs://openhands-evaluation-results/metadata/*.jsonl`), so an in-flight eval against a >2mo SDK sha could lose its image mid-run. Low probability given the 100-version buffer; proper fix (pre-delete jsonl filter) is a follow-up PR. cc @juanmichelini @simonrosenberg.

## Test plan

- [ ] Add `GHCR_CLEANUP_PAT` secret
- [ ] Confirm package-repo link has `Write` role
- [ ] Manual dispatch with `dry-run: true` — inspect the deletion list
- [ ] Manual dispatch with `dry-run: false` for the first real purge
- [ ] Verify post-run tag count via `ghcr.io/v2/openhands/eval-agent-server/tags/list` (expect ~130k → a few thousand)
- [ ] Confirm a subsequent `build-swebench-images.yml` run still succeeds

### DIFF
--- a/.github/workflows/ghcr-retention.yml
+++ b/.github/workflows/ghcr-retention.yml
@@ -36,11 +36,9 @@ jobs:
   clean:
     runs-on: ubuntu-latest
     timeout-minutes: 720  # 12h — first real run may delete tens of thousands of versions
-    permissions:
-      packages: write
     steps:
       - name: Delete old eval-agent-server versions
-        uses: snok/container-retention-policy@v3.0.0
+        uses: snok/container-retention-policy@4f22ef80902ad409ed55a99dc5133cc1250a0d03  # v3.0.0
         with:
           account: OpenHands
           token: ${{ secrets.GHCR_CLEANUP_PAT }}
--- a/.github/workflows/ghcr-retention.yml
+++ b/.github/workflows/ghcr-retention.yml
@@ -1,0 +1,52 @@
+name: GHCR retention - eval-agent-server
+
+# Nightly garbage collection of ghcr.io/openhands/eval-agent-server.
+# Context: see issue #684 — the package had 133k+ tags with no retention,
+# fanning out per SDK sha (97.9% of tags). This job deletes versions older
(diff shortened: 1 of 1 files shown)


juanmichelini (approved):
LGTM, this is a GitHub action, approving so it can be tested

simonrosenberg:
> LGTM, this is a GitHub action, approving so it can be tested

we can always test it by overriding another pre-existing github action and then renaming it before merging btw!

Issue #576, opened 2026-03-26
Tracking: ACP Agent Full Benchmark Runs


simonrosenberg:
## ACP Agent End-to-End Benchmark Runs

Comprehensive tracking of all **full** benchmark runs using ACP agents (`acp-claude`, `acp-codex`, and `acp-gemini`).

**Full run thresholds:** swebench=500, swtbench=433, gaia=165, commit0=16, swebenchmultimodal=68

---

## Master Table (latest run per model per benchmark)

> Validated with [`validate_artifacts.py`](<https://github.com/OpenHands/benchmarks/blob/main/validate_artifacts.py>) — checks: archive readable, logs exist, coverage >= 50%. Cost telemetry checked for acp-claude only (unavailable for acp-codex, proxy-reported for acp-gemini).

### Codex GPT-5.4

| Benchmark | Run ID | Coverage | Resolved | Proxy Cost | Validation | GCS Artifacts |
| -- | -- | -- | -- | -- | -- | -- |
| swebench | [23968384393](<https://github.com/OpenHands/evaluation/actions/runs/23968384393>) | 497/500 (99.4%) | 369 | $197.86 | **PASS** | [results.tar.gz](<https://results.eval.all-hands.dev/swebench/litellm_proxy-openai-gpt-5-4/23968384393/results.tar.gz>) |
| swtbench | [23983221678](<https://github.com/OpenHands/evaluation/actions/runs/23983221678>) | 428/430 (99.5%) | 314 | $150.95 | **PASS** | [results.tar.gz](<https://results.eval.all-hands.dev/swtbench/litellm_proxy-openai-gpt-5-4/23983221678/results.tar.gz>) |
| gaia | [24015998468](<https://github.com/OpenHands/evaluation/actions/runs/24015998468>) | 165/165 (100%) | 140 | $36.35 | **PASS** | [results.tar.gz](<https://results.eval.all-hands.dev/gaia/litellm_proxy-openai-gpt-5-4/24015998468/results.tar.gz>) |
| commit0 | [23968384476](<https://github.com/OpenHands/evaluation/actions/runs/23968384476>) | 16/16 (100%) | 10 | $13.94 | **PASS** | [results.tar.gz](<https://results.eval.all-hands.dev/commit0/litellm_proxy-openai-gpt-5-4/23968384476/results.tar.gz>) |
[...]

PR #700, opened 2026-04-26
fix(commit0): increase conversation and instance timeouts to 4h

simonrosenberg:
## Summary

- Adds `CONVERSATION_TIMEOUT = 4 * 60 * 60` (4h) to `commit0/config.py`, replacing the implicit 3600s default
- Adds `INSTANCE_TIMEOUT = 4 * 60 * 60` (4h) and passes it to `Evaluation.__init__()` so the per-instance hard kill matches the conversation budget
- Wires `CONVERSATION_TIMEOUT` through to both `acp_prompt_timeout` in `build_acp_agent` and `conversation.run(timeout=...)` — both previously hard-coded to 3600s
- Adds an optional `prompt_timeout` param to `build_acp_agent` so benchmarks can override the global ACP default without touching `acp.py` constants

## Motivation

Observed during Opus 4.7 commit0 run: babel, chardet, and marshmallow all failed with `Remote conversation ended with error` after exactly 3600s on every retry. Root cause: `CONVERSATION_TIMEOUT` and `acp_prompt_timeout` both defaulted to 3600s, and these large repos need more time.

Data from live runs:
- marshmallow: ~98 min
- chardet: ~60 min
- babel: >3h

Both `CONVERSATION_TIMEOUT` and `INSTANCE_TIMEOUT` are set to 4h so they are aligned — the conversation timeout fires cleanly before the hard kill.

### DIFF
--- a/benchmarks/commit0/config.py
+++ b/benchmarks/commit0/config.py
@@ -15,11 +15,10 @@ CONDENSER_DEFAULTS = {
 
 # Timeout for each conversation run in seconds.
 # marshmallow: ~98 min, chardet: ~60 min, babel: >3h (exhausted 3h limit without committing)
-# Using 6h (21600s) for babel which needs extended exploration of a large codebase.
-CONVERSATION_TIMEOUT = 21600
+CONVERSATION_TIMEOUT = 4 * 60 * 60
 
 # Per-instance hard kill timeout in seconds (passed to Evaluation base class).
-# Set to 4h to match CONVERSATION_TIMEOUT headroom; babel needs >3h of active work.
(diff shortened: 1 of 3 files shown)


juanmichelini (commented):
@simonrosenberg merging this one would break 1 to 1 comparison with already submitted result.

simonrosenberg:
> @simonrosenberg merging this one would break 1 to 1 comparison with already submitted result.

I guess the question is: for results in openhands-index, how many commit0 inference timeout errors? 
If there aren't any it doesn't break 1 to 1 comparison, let me investigate

simonrosenberg:
Closing per reviewer feedback — merging would break 1-to-1 comparison with already submitted results. The timeout issue is documented in #576 Known Issues instead.

PR #701, opened 2026-05-01
fix(swebench): explicitly close conversation after evaluate_instance to prevent OOM

simonrosenberg:
## Summary

- Wrap the conversation body in `try/finally` and call `conversation.close()` before returning from `evaluate_instance()` in both `swebench` and `swebenchmultimodal` `run_infer.py`
- `close()` sets `self._ws_client = None`, breaking the circular reference so CPython's reference counter can immediately reclaim `_cached_events`

## Root cause (OpenHands/evaluation#540)

`RemoteConversation` has a circular reference that prevents immediate GC:

```
conversation → _ws_client (WebSocketCallbackClient)
             → callback (composed_callback)
             → run_complete_callback closure (remote_conversation.py:813)
             → self._terminal_status_queue   ← captures self
```

Without an explicit `close()` call, CPython's cyclic GC must collect these objects, and it only runs on a threshold — not immediately. With 30 workers completing instances rapidly, many `RemoteConversation` objects pile up.

This is especially bad for ACP agents (`acp-codex`, `acp-claude`) because:
- `ACPAgent` raises `NotImplementedError` if a condenser is configured, so all events accumulate uncapped in `_cached_events`
- GPT-5.5 with `reasoning_effort=high` on large JS projects (wp-calypso in swebenchmultimodal) makes 100+ tool calls per instance; each `ACPToolCallEvent` stores `raw_input` + `raw_output` which can be MB-sized for file reads and test output

**Observed failure**: swebenchmultimodal with `acp-codex` OOMed the eval-container (12Gi limit) at ~40/68 instances across 5 consecutive runs. Both `num_workers=8` and `num_workers=30` hit the same fraction — confirming per-instance linear accumulation rather than concurrency pressure.

## Test plan

- [ ] Trigger a swebenchmultimodal eval with `acp-codex` agent type and verify it runs to completion without OOMKill
- [ ] Verify swebench eval still works normally (no regression)
- [ ] Confirm the fix applies to both files identically

### DIFF
--- a/benchmarks/swebench/run_infer.py
+++ b/benchmarks/swebench/run_infer.py
@@ -330,74 +330,84 @@ class SWEBenchEvaluation(Evaluation):
             delete_on_close=True,
         )
 
-        logger.info("repo_path: %s", repo_path)
-        cp_testebed_repo = workspace.execute_command(
-            (f"mkdir -p {repo_path} ; cp -r /testbed/. {repo_path}")
-        )
-        assert cp_testebed_repo.exit_code == 0, (
-            f"cp_testebed_repo failed: {cp_testebed_repo.stderr}"
-        )
+        try:
+            logger.info("repo_path: %s", repo_path)
+            cp_testebed_repo = workspace.execute_command(
+                (f"mkdir -p {repo_path} ; cp -r /testbed/. {repo_path}")
+            )
+            assert cp_testebed_repo.exit_code == 0, (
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Closing in favor of a cleaner fix: gc.collect() in the Evaluation base class (one change covers all benchmarks) + weakref fix in the SDK to eliminate the circular reference at the source.

PR #702, opened 2026-05-01
fix(evaluation): gc.collect() after each instance to reclaim cyclic RemoteConversation objects

simonrosenberg:
## Summary

Add `gc.collect()` after each instance result is persisted, so the cyclic GC runs once per instance rather than on an unpredictable threshold.

## Why

`RemoteConversation` has a circular reference via its WebSocket callback closure that prevents CPython's reference counter from freeing it immediately. Objects with large `_cached_events` lists accumulate across instances and can OOM the eval-container. Fixes OpenHands/evaluation#540.

## Companion PR

OpenHands/software-agent-sdk#3031 eliminates the cycle at the source with a `weakref`. This `gc.collect()` is a belt-and-suspenders safety net.

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -650,13 +650,9 @@ class Evaluation(ABC, BaseModel):
                 # from the attempt files, not from this in-memory list.
                 out.history = []
 
-                # Force the cyclic GC to run now. RemoteConversation objects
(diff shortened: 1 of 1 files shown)


simonrosenberg:
Closing because not needed, this fixes it https://github.com/OpenHands/software-agent-sdk/pull/3033

Issue #708, opened 2026-05-12
swtbench: qwen3-coder-next score is artificially low — agent writes source-code fix alongside the test (78% of patches)


juanmichelini:
## Summary

On the recent `swtbench` run [`litellm_proxy-openrouter-qwen-qwen3-coder-next/24103435463`](https://results.eval.all-hands.dev/swtbench/litellm_proxy-openrouter-qwen-qwen3-coder-next/24103435463/results.tar.gz) the model resolves **14 / 424 (3.3 %)** instances ([monitor link](https://openhands-eval-monitor.vercel.app/?days=90&benchmark=swtbench&text=qwen&run=swtbench%2Flitellm_proxy-openrouter-qwen-qwen3-coder-next%2F24103435463)). Analyzing the predicted patches shows the score is almost entirely explained by the agent modifying *source code* alongside its test, which destroys the F2P signal SWT-bench scores on.

## Evidence

Classifying every patch in `output.swtbench.jsonl` by which kinds of files it touches and cross-tabbing with `output.report.json`:

| Patch category                              | # instances | Resolved | Rate      |
| ------------------------------------------- | ----------: | -------: | --------- |
| **Only test files** (clean)                 |          22 |       10 | **45.5 %** |
| Mixed: test files **and** source code       |         332 |        4 | 1.2 %     |
| Only source code / no test                  |          64 |        0 | 0 %       |
| Empty                                       |           3 |        0 | 0 %       |
| **Total**                                   |     **424** |   **14** | **3.3 %** |

Two observations:

1. The agent's tests are actually fine when left alone — **45 %** solve rate on the "only test files" subset, comparable to what we'd expect for this model class.
2. The same agent's solve rate drops to **1.2 %** as soon as the patch also touches source. That's the expected SWT-bench failure mode: if the model's patch also "fixes" the bug, its own test no longer fails on `model_patch`-applied state, so no F2P credit.

**78 % of patches (332 / 424)** fall into the "mixed" bucket — that is what is dragging the score down.

### What the extra (non-test) changes look like

Across the 424 patches the non-test files modified are:

```
  814  source code (e.g. sympy/core/basic.py, django/db/models/sql/query.py, …)
   65  build/lib/...                              (1 instance, psf__requests-1142)
   18  docs/ or doc/
    7  reproduction.py / reproduction_script.py   (repo root)
    1  FIX_SUMMARY.md
    +  15 patches also contain root-level scratch test_*.py files alongside real tests
```

[...]

PR #710, opened 2026-05-13
DRAFT: swtbench: tighten default prompt to discourage non-test edits (option 1 of #708)

juanmichelini:
Addresses option 1 from #708.

## What

Tightens the existing SWT-bench prompt at `benchmarks/swtbench/prompts/default.j2` so the agent stops modifying source code and scratch files alongside its tests.

The previous default prompt was self-contradictory:

* it said _"DON'T have to modify the actual logic and ONLY have to update test logic and tests"_…
* …but then step 2 said _"Create a script `reproduction.py`"_ and step 3 said _"Edit the sourcecode of the repo to integrate your reproduction script into the test framework"_.

That's a lot of the failure mode on qwen3-coder-next (#708 found 78% of patches touch source or scratch files; the test-only subset solves at 45.5%, mixed at 1.2%).

This PR edits `default.j2` to:

- add an `IMPORTANT` block stating exactly which paths are scored and *why* touching source files silences the F2P signal;
- ban scratch files at the repo root (`reproduction.py`, `test_repro.py`, `FIX_SUMMARY.md`), and `build/`, `docs/`, `pyproject.toml`, etc.;
- replace step 2 (`create reproduction.py`) with "use BashTool to confirm the buggy behavior";
- replace step 3 (`edit sourcecode`) with "edit only existing test files inside the test directory";
- add a final step: run `git diff --name-only` and revert anything outside the test directory.

Diff is 13 added / 4 removed lines, no new files.

## Scope

- Single file changed: `benchmarks/swtbench/prompts/default.j2`.
- Applies to every SWT-bench run that doesn't pass `--prompt-path` explicitly — including the SDK [`run-eval.yml`](https://github.com/OpenHands/software-agent-sdk/actions/workflows/run-eval.yml), which doesn't thread `--prompt-path` through, so the change activates automatically once merged.

## How to test

Re-run any SWT-bench config — no extra flags needed. Compare the resulting `output.swtbench.jsonl` patch-shape distribution to the baseline run linked in #708: the "mixed test + source" share should drop substantially. Score lift only materializes if the agent actually follows the instructions; if it doesn't, #711 (post-processing strip) is the safety net.

## Risks / caveats

- This is a behavior nudge to *every* model running SWT-bench, not just qwen3-coder-next. Models that were already test-only are unaffected by the new rules. Models that produced "useful" source-code fixes for some other reason would lose them — but those edits aren't scored anyway, so the only downside is the agent might try harder to comply and waste more turns on test edits. Acceptable IMO; happy to revisit if a strong model regresses.
- Wording is intentionally direct ("DO NOT", "earns zero credit"); if you'd rather keep the tone softer let me know and I'll soften it.
- Does NOT address the agent that ignores instructions entirely — that case is what #711 handles.

Refs #708. Supersedes the earlier version of this PR that added a separate `qwen3_coder_next.j2` template.

---

_This PR was created by an AI agent (OpenHands) on behalf of @juanmichelini._

### DIFF
--- a/benchmarks/swtbench/prompts/default.j2
+++ b/benchmarks/swtbench/prompts/default.j2
@@ -10,10 +10,19 @@ I've uploaded a python code repository in the directory {{ workspace_dir_name }}
 
 Can you help me implement the necessary changes to the repository to test whether the issue in <issue_description> was resolved?
 I will take care of all changes to any of the non-test files. This means you DON'T have to modify the actual logic and ONLY have to update test logic and tests!
-Your task is to make the minimal changes to tests files in the /workspace directory to reproduce the issue in the <issue_description>, i.e., such that the generated tests fail in the current state (where the issue is unresolved) and pass when the issue will be resolved.
+Your task is to make the minimal changes to existing test files in the /workspace directory to reproduce the issue in the <issue_description>, i.e., such that the generated tests fail in the current state (where the issue is unresolved) and pass when the issue will be resolved.
+
+IMPORTANT — only the diff against existing test files inside the repository's test directory (e.g. `tests/`, `test/`, `testing/`) will be scored:
+* DO NOT modify any source file (e.g. files under `sympy/`, `django/`, `sphinx/`, `requests/`, …). Doing so makes your test pass against your own "fix" instead of failing on the buggy code, which silences the bug-reveal signal and earns zero credit.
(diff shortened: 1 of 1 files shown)


juanmichelini:
Tests show tightnening prompt does not work for certain models

PR #712, opened 2026-05-14
swtbench: don't lose a whole run when upstream grading.py crashes on one bad log

juanmichelini:
## What

Patches the cloned upstream `swt-bench/src/grading.py` from `ensure_swt_bench_repo` so a single malformed test log can't sink an entire benchmark report.

## Why

In run [25868145876](https://github.com/OpenHands/evaluation/actions/runs/25868145876) ([eval monitor](https://openhands-eval-monitor.vercel.app/?run=swtbench%2Flitellm_proxy-trinity-large-thinking%2F25868145876%2F&days=15)), inference completed cleanly (`100%|██████████| 432/432`), but the reporting pass crashed at instance ~236/432 with:

```
File "/root/.cache/openhands/swt-bench/swt-bench/src/grading.py", line 76, in get_logs_eval
    content = re.split(r"\n\+ python3 [^\n]*trace.py --count -C coverage.cover [^\n]*\n",
                       raw_content, flags=re.MULTILINE)[1]
IndexError: list index out of range
```

Upstream code:

```python
if "trace.py --count -C coverage.cover" in raw_content:
    content = re.split(
        r"\n\+ python3 [^\n]*trace.py --count -C coverage.cover [^\n]*\n",
        raw_content, flags=re.MULTILINE,
    )[1]
else:
    content = raw_content
```

The substring guard is **strictly weaker** than the regex it gates: a log can contain the substring `trace.py --count -C coverage.cover` without containing the full shell-trace marker line `\n+ python3 … trace.py --count -C coverage.cover …\n` (e.g. when the test container died or got truncated mid-run, or the prefix is `+ python` instead of `+ python3`). `re.split` then returns a single-element list and `[1]` raises `IndexError`. Because the exception is uncaught, it propagates `get_logs_eval → report_results → make_run_report → src/main.py`, so **one bad log aborts the report for all 432 instances** and `report.json` is never written. The whole 2h+ run is lost.

## How

Two idempotent in-place patches applied right after `git clone` in `ensure_swt_bench_repo`:

1. **`get_logs_eval` defensive split** — fall back to `content = raw_content` when `re.split(...)` returns fewer than two chunks. Same behavior as the existing `else` branch.

2. **`report_results` per-instance isolation** — wrap the function so any uncaught grading exception is logged with a full traceback and the instance is reported as `{"resolved": False, "coverage_pred": None, ...}`. This is belt-and-suspenders: even if a future grading bug shows up in some other code path, the run still produces a report.

Both helpers:
- Are guarded by a marker comment → safe to re-apply on a cached clone.
- No-op with a warning if the upstream snippet shape changes, so we never wedge the run if upstream rewrites that code.

## Tests

New `tests/test_swtbench_grading_patches.py` (5 tests, all hermetic — no network/docker):

- `test_unpatched_get_logs_eval_raises_index_error` — reproduces the exact production bug against a minimal fake upstream layout.
- `test_patched_get_logs_eval_falls_back_to_raw_content` — same input no longer raises.
- `test_patches_are_idempotent` — applying twice produces identical bytes.
- `test_patched_report_results_isolates_failures` — simulated `RuntimeError` inside the wrapped function returns a synthetic unresolved dict instead of propagating.
- `test_patch_helpers_no_op_when_grading_missing` — defensive against unexpected upstream layout.

```
tests/test_swtbench_grading_patches.py::test_unpatched_get_logs_eval_raises_index_error PASSED
tests/test_swtbench_grading_patches.py::test_patched_get_logs_eval_falls_back_to_raw_content PASSED
tests/test_swtbench_grading_patches.py::test_patches_are_idempotent PASSED
tests/test_swtbench_grading_patches.py::test_patched_report_results_isolates_failures PASSED
tests/test_swtbench_grading_patches.py::test_patch_helpers_no_op_when_grading_missing PASSED
============================== 5 passed in 0.04s ==============================
```

Pre-commit (ruff format/lint, pyright strict) clean on changed files.

## Comparability with historical results

This change does not alter the score for any instance whose log was well-formed — `get_logs_eval`'s control flow for valid logs is byte-for-byte unchanged (`_parts[1]` is taken when the regex matches, exactly as before). It only changes behavior for instances that previously caused the entire harness to crash. Those instances now score "unresolved" (which is the truthful outcome — their test container failed to produce a valid log) instead of aborting the run.

---

_This PR was created by an AI agent (OpenHands) on behalf of @juanmichelini after investigating the failed run together._

@juanmichelini can click here to [continue refining the PR](https://app.all-hands.dev/conversations/cad590eb-a849-465d-aef9-e340346229a1)

### DIFF
--- a/benchmarks/swtbench/image_utils.py
+++ b/benchmarks/swtbench/image_utils.py
@@ -15,6 +15,132 @@ from openhands.sdk import get_logger
 logger = get_logger(__name__)
 
 
+_GRADING_PATCH_MARKER = "# openhands-benchmarks: defensive coverage-marker split"
+_REPORT_ISOLATION_MARKER = "# openhands-benchmarks: per-instance report isolation"
+
+
+def _patch_grading_get_logs_eval(swt_bench_dir: Path) -> None:
+    """
+    Make upstream ``src/grading.py::get_logs_eval`` defensive against missing
+    coverage-trace markers.
+
+    Upstream code:
+
+        if "trace.py --count -C coverage.cover" in raw_content:
+            content = re.split(
+                r"\\n\\+ python3 [^\\n]*trace.py --count -C coverage.cover [^\\n]*\\n",
+                raw_content,
+                flags=re.MULTILINE,
+            )[1]
+
+    The substring guard is weaker than the regex (different line prefix,
+    truncated logs, etc.), so ``re.split(...)`` can return a single-element
+    list and ``[1]`` raises ``IndexError``. A single bad instance log then
+    aborts ``make_run_report`` for the entire run.
+
+    This patch falls back to the full ``raw_content`` when the marker line
+    is not present, matching the ``else`` branch's behavior. Idempotent.
+    """
+    grading_py = swt_bench_dir / "src" / "grading.py"
+    if not grading_py.exists():
+        logger.warning(
+            "Cannot patch swt-bench grading.py: %s does not exist", grading_py
+        )
+        return
+
+    text = grading_py.read_text(encoding="utf-8")
+    if _GRADING_PATCH_MARKER in text:
+        return
+
+    old = (
+        '        content = re.split(r"\\n\\+ python3 [^\\n]*trace.py --count -C '
+        'coverage.cover [^\\n]*\\n", raw_content, flags=re.MULTILINE)[1]\n'
+    )
+    new = (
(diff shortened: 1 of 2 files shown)


juanmichelini:
testing here https://openhands-eval-monitor.vercel.app/?run=swtbench%2Flitellm_proxy-trinity-large-thinking%2F25882238007%2F&days=15

PR #780, opened 2026-09-04
fix(eval): wait for LiteLLM spend to settle before recording proxy_cost

simonrosenberg:
## Why

`proxy_cost` is the exact per-instance spend we report from LiteLLM virtual keys, and it is the number the ACP harness comparison (OpenHands/software-agent-sdk#4627, #4638) will be built on. It is currently biased low.

LiteLLM commits virtual-key spend **asynchronously**, a few seconds after each request returns. `_query_proxy_cost` runs immediately after `evaluate_instance()` — right inside that window — and its retry logic is looking for the wrong thing:

```python
proxy_cost = get_key_spend(virtual_key)
if proxy_cost is None or proxy_cost == 0.0:
    for delay in (2, 4, 8, 16):
        time.sleep(delay)
        retry_cost = get_key_spend(virtual_key)
        if retry_cost is not None and retry_cost > 0:
            proxy_cost = retry_cost
            break          # <- breaks on the first partial read
```

It retries only while spend is exactly `$0`, then `break`s on the first value above zero. A read that is non-zero but still climbing is accepted as final, so the ladder actively *selects* the partial value.

## Evidence

Measured against `llm-proxy.eval.all-hands.dev` by polling `/key/info` at fixed offsets after a single ACP turn ended (two harnesses, both reproducible):

| offset after run | opencode | kimi |
|---|---|---|
| t+0s | $0 | $0 |
| t+2s | $0 | $0 |
| **t+5s** | **$0.01308805** | **$0.02536200** |
| t+10s | $0.01541885 | $0.07714400 |
| t+20s … t+95s | $0.01541885 | $0.07714400 |

Replaying both algorithms over those curves:

| harness | settled truth | current code | error | this PR |
|---|---|---|---|---|
| opencode | $0.01541885 | $0.01308805 @ t+6s | **-15.1%** | $0.01541885 @ t+15s |
| kimi | $0.07714400 | $0.02536200 @ t+6s | **-67.1%** | $0.07714400 @ t+15s |

The size of the error depends on how much of the final total the last few requests carry, which varies per harness — so this is not a constant offset a paired comparison can absorb.

This is pre-existing and independent of which harness is used: `acp-claude` and `acp-codex` are affected today.

## Summary

- Poll until **two consecutive reads agree** rather than until spend is non-zero.
- Fixed 5s interval, 12 attempts (60s ceiling), matching the 5–10s commit latency measured above. The old exponential ladder was shaped to wait for spend to *appear*, which is no longer the goal.
- `$0` still warns as before. A total that never settles now warns too, instead of being reported silently as if it were exact.

Worst case adds ~30s per instance versus the old path; instances run concurrently in worker threads, so this is wall-clock noise against a benchmark run.

## How to Test

```bash
uv run --frozen pytest tests/test_evaluation_proxy_cost.py -q   # 6 passed
uv run --frozen ruff check benchmarks/utils/evaluation.py tests/test_evaluation_proxy_cost.py
```

`test_proxy_cost_waits_for_partially_committed_spend_to_settle` is the regression test — it feeds the measured kimi sequence `[0.0, 0.025362, 0.077144, 0.077144]` and asserts `0.077144`. The old code returns `0.025362`.

Full suite: `547 passed, 13 failed`. All 13 failures are in `tests/test_programbench.py` (hook scripts) and reproduce identically on a clean `origin/main` with this change stashed.

### DIFF
--- a/benchmarks/utils/evaluation.py
+++ b/benchmarks/utils/evaluation.py
@@ -59,6 +59,12 @@ from openhands.workspace import APIRemoteWorkspace
 
 logger = get_logger(__name__)
 
+# LiteLLM commits virtual-key spend asynchronously; a read taken right after the
+# last request can catch a partially committed total. Measured commit latency is
+# 5-10s, so poll on that scale and cap the wait at one minute.
+_PROXY_SPEND_POLL_INTERVAL = 5
+_PROXY_SPEND_POLL_ATTEMPTS = 12
+
 # Interval in seconds between checking for per-instance timeouts
 TIMEOUT_CHECK_INTERVAL_SECONDS = 60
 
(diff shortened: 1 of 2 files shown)


juanmichelini (approved):
LGTM

